PR #35, opened 2026-04-23
fix(team-management): prevent invite-link overflow in modal (#30)

### LINKED ISSUE #30: UI Issue - Form Field Overflows Parent Modal
To reproduce: 
Navigate to http://localhost:5173/settings/teams
Pick a team and attempt to add a member

You'll see that the "Invitation created for john.doe@precogly.dev. Share this link with them:" form field overflows the parent "Manage Team: My Team" modal.

Screenshot below

Image

mvanhorn:
Closes #30.

The "Invitation created for ... Share this link with them:" panel in `frontend/src/features/organization/pages/TeamManagement.tsx` used `flex-1` and `truncate` on the inner `<code>` without the `min-w-0` + `block overflow-hidden` combo that Tailwind's truncate needs to actually clip in a flex container. With a long invite URL, the code element expanded past the modal's right edge — the screenshot in the issue.

The same panel in `frontend/src/features/threat-models/components/workspace/ManagePeopleModal.tsx` already has the fix (lines 208–209). This PR just mirrors that pattern in TeamManagement.tsx so both invite panels behave the same way.

## Diff

```diff
-                    <div className="flex items-center gap-2">
-                      <code className="flex-1 text-xs bg-background p-2 rounded border truncate">
+                    <div className="flex items-center gap-2 min-w-0">
+                      <code className="flex-1 text-xs bg-background p-2 rounded border truncate block overflow-hidden">
```

Two-attribute change, no behavior or visual changes for short URLs.

### DIFF
--- a/frontend/src/features/organization/pages/TeamManagement.tsx
+++ b/frontend/src/features/organization/pages/TeamManagement.tsx
@@ -371,9 +371,9 @@ function TeamMembersDialog({ team, open, onOpenChange }: TeamMembersDialogProps)
           </DialogDescription>
         </DialogHeader>
 
-        <div className="space-y-4">
+        <div className="space-y-4 min-w-0">
           {/* Invite form */}
-          <div className="space-y-2">
+          <div className="space-y-2 min-w-0">
             <Label>Invite Member</Label>
             <div className="flex gap-2">
               <Input
@@ -381,7 +381,7 @@ function TeamMembersDialog({ team, open, onOpenChange }: TeamMembersDialogProps)
                 placeholder="Email address"
                 value={inviteEmail}
                 onChange={(e) => setInviteEmail(e.target.value)}
-                className="flex-1"
+                className="flex-1 min-w-0"
               />
               <Select
                 value={inviteRole}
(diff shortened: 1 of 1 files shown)


mvanhorn:
Heads up - the cla-assistant.io link in the CLA check returns "Bad credentials" when I try to sign, so I'm stuck here. Happy to sign once the CLA integration is working, or let me know if there's another path.

vikram-s-narayan:
@mvanhorn - thanks for the contribution and sorry for your challenges. I'd inadvertently removed CLA bot from GitHub authorized apps. Could you please try again?

mvanhorn:
Signed now - thanks for fixing the bot, @vikram-s-narayan!

vikram-s-narayan:
@mvanhorn - thank you for this PR. It looks like the issue is not fixed ... at least on my system.
Could you please check if it is working on your system?

Screenshot on my system from the PR is below:
Screenshot 2026-04-26 at 6 53 43 AM

mvanhorn:
Thanks @vikram-s-narayan, dug into why the previous fix wasn't holding up on your end and pushed ad2bcee.

The root cause: DialogContent uses CSS grid, and grid items have `min-width: auto` by default. That means even with `min-w-0` on the immediate flex container of the invite link, the long URL was still able to push the grid child (and therefore the whole dialog) wider than the viewport.

The new commit propagates `min-w-0` along the chain: the grid child div, the invite-form section, the muted result box, and the inner wrapper. Once each ancestor permits shrinking, the flex item can actually be constrained.

I also switched the `<code>` from `truncate` to `break-all` so the long invite URL wraps onto multiple lines inside the muted box instead of being clipped. Easier to read and select if the copy-to-clipboard button isn't enough. Added `shrink-0` to the copy Button so it isn't squeezed.

Verified locally: `tsc -b` is clean for TeamManagement.tsx (the build's other errors are pre-existing on main).

If you'd prefer the URL stays on one line and the dialog grows a scrollbar instead of wrapping, happy to switch the `code` back to `truncate` and add `overflow-x-auto` instead - just let me know.

vikram-s-narayan:
Hi @mvanhorn - I will test this tomorrow on my system and will merge. Thank you!!!

vikram-s-narayan:
This looks great! It really does hold up even if we change the window size. I also tested out with Safari and it works perfectly. Merging now.
Screenshot 2026-04-28 at 6 32 09 AM

mvanhorn:
Thanks @vikram-s-narayan, glad the grid fix held up. Appreciate you testing across browsers.

Issue #40, opened 2026-04-25
Show/Hide Password Toggle on Login Screen


skotreshi:
It would be a useful enhancement to add a show/hide password option on the login screen.

Currently, if a user enters an incorrect character, it’s difficult to identify the mistake, which can lead to repeated login attempts. A toggle to reveal the password would improve usability and reduce friction.

PR #64, opened 2026-05-08
feat(auth): add show/hide password toggle to login and signup forms

BhavyaB19:
Implements show/hide password toggle on login and signup pages as outlined in #40

**Changes Made**
- Added password visibility toggle for login password field
- Added independent password visibility toggles for signup password and confirm password fields
- Updated input types dynamically between `password` and `text`
- Added accessible labels for toggle buttons using aria-label

**Testing**

- Verified password visibility toggles work correctly
- Confirmed each signup field toggle works independently
- Checked login flows after changes
- Verified no TypeScript or build errors

https://github.com/user-attachments/assets/5bebc430-5d7c-464f-8528-45214ed15188


vikram-s-narayan:
Thanks @BhavyaB19!

Issue #56, opened 2026-05-03
Add Delete for Components, Threats and Countermeasures in Threat Analysis View


vikram-s-narayan:
## Summary

The Threat Analysis view currently has no way to permanently delete components, threats, or countermeasures. This issue adds delete functionality for all three entity types.

## Current Behavior

- **Components:** Can be added but not removed
- **Threats:** Can be dismissed (soft delete) and restored, but not permanently deleted
- **Countermeasures:** Can be added but not removed (despite backend + frontend mutations already existing)

### 1. Countermeasure Delete (easiest, start here)

Add a trash icon button to each countermeasure card. On click, show a confirmation dialog, then call the existing delete mutation.

**Key files:**
- `frontend/src/features/dfd-editor/components/threat-analysis/ComponentView.tsx` - Countermeasure cards render around line ~1012. Add a `Trash2` icon button in the card header, visible on hover (the card already has `group` class for hover state)
- `frontend/src/features/threat-models/api/threats.ts` - **Mutations already exist:** `useDeleteCountermeasure()` (line ~522) and `useDeleteFlowCountermeasure()` (line ~538). No API work needed
- Use `parseCountermeasureId()` (line ~560 in threats.ts) to determine whether to call the component or flow variant

**UI pattern to follow:** See the existing trash icon in `EditComplianceMappingsDialog.tsx` (line ~252):
```tsx
<Button
  variant="ghost"
  size="icon"
  className="h-8 w-8 text-destructive hover:text-destructive"
  onClick={() => handleDelete(id)}
>
  <Trash2 className="h-4 w-4" />
</Button>
```

### 2. Threat Delete (two-step UX)

Threats use a dismiss/restore workflow. To prevent accidental permanent deletion, the delete button should only appear in the **Dismissed** section, next to the existing "Restore" button.

**UX flow:** Dismiss (X button) -> threat moves to Dismissed section -> user can Restore or permanently Delete (trash icon)

**Key files:**
- `frontend/src/features/dfd-editor/components/threat-analysis/ComponentView.tsx` - Dismissed threats render at line ~899. Add a `Trash2` button next to the existing "Restore" button (line ~914)
- `frontend/src/features/threat-models/api/threats.ts` - **Need to add** `useDeleteThreat()` and `useDeleteFlowThreat()` mutations. Follow the same pattern as `useDeleteCountermeasure()`:

```typescript
export function useDeleteThreat() {
  const queryClient = useQueryClient()
  return useMutation({
[...]

Issue #2, opened 2026-04-14
Threats and Countermeasures Should Show up in DFD Editor


vikram-s-narayan:
Currently, when users add in a component in the DFD editor, they need to then go to the threat analysis workspace in order to add threats and countermeasures. Ideally, they should be able to add threats and countermeasures on components _inside_ the DFD editor. Additionally, if a threat or countermeasure is added in the threat analysis workspace, it should reflect the update inside the DFD editor too.

What the DFD editor shows per node: label, technology, data
  sensitivity badge. That's it.

  What exists in the database but is invisible on the DFD:
  - Threat count per component
  - Threat severity (inherent + residual)
  - Threat status (exposed/addressable/mitigated)
  - Countermeasure status (gap/planned/verified/waived/platform)
  - Countermeasure ownership (assigned_owner)
  - Inherited countermeasures from zones

  So when you add threats and countermeasures to "Amazon S3" in the
  threat analysis workspace, the S3 node on the DFD canvas stays
  completely unchanged. The data is in the database, linked via
  OrgsystemComponent, but the DFD editor never fetches or renders it.

  The two screens are essentially operating on the same underlying data
  through separate, non-communicating UIs.

But implementing this fix requires us to address questions like what happens if a user deletes a node in the threat analysis workspace?

Screenshots added for additional reference.

<img width="1502" height="790" alt="Image" src="https://github.com/user-attachments/assets/afea11ab-8a3b-48e9-9e76-027ec7ed01cf" />
<img width="1504" height="781" alt="Image" src="https://github.com/user-attachments/assets/e595b47c-cbfe-4f15-bd44-9687ed5b2ee7" />

sidd190:
Helloo, I'm working on this! Can you please assign it to me! It's almost working locally, I'll take a day or two to get a PR for this.

vikram-s-narayan:
Oh wow! that's really cool. It's a monster feature with lots of moving parts. Take as much time as you need! Assigned the task to you.

sidd190:
Awesome! Thanks!

vikram-s-narayan:
Just adding this discussion link to this issue for further reference - https://threatmodelingconnect.discourse.group/t/vicar-bridging-the-gap-between-threat-modeling-and-remediation/1003/5

sidd190:
[...]

PR #79, opened 2026-05-16
feat: add delete functionality for components, threats and countermeasures

BhavyaB19:
Implements Add Delete for Components, Threats and Countermeasures in Threat Analysis View as outlined in #56 

**Changes Made**
- Added a hover-only trash icon to each component in the left sidebar.
- Added a hover-only trash icon to dismissed threats.
- Added a hover-only trash icon to each countermeasure card.
- Wired each delete action to the existing API mutations for countermeasures.
- Added new API mutations for threats and components - `useDeletethreat()`, `useDeleteFlowThreat()` and `useDeleteComponent()`
- Added confirmation dialogs for all three delete flows.
- Updated the component delete confirmation copy to warn that deleting a component also removes its associated threats and countermeasures.
- Kept the styling consistent with the existing destructive action pattern using the shared destructive color token.

**Behaviour**
- Delete icons only appear on hover to reduce visual noise.
- Clicking a trash icon opens a confirmation dialog instead of deleting immediately.
- Confirming a component delete cascades through related threats and countermeasures as supported by the backend.
- Threat and countermeasure deletes use their existing flow/component-aware mutations.

https://github.com/user-attachments/assets/f77ec71f-0e45-4d80-9b25-c35e385e10a0

### DIFF
--- a/frontend/src/features/dfd-editor/components/threat-analysis/ComponentTreeItem.tsx
+++ b/frontend/src/features/dfd-editor/components/threat-analysis/ComponentTreeItem.tsx
@@ -1,5 +1,6 @@
-import { Cog, Database, User, ChevronRight, Building2 } from 'lucide-react'
+import { Cog, Database, User, ChevronRight, Building2, Trash2 } from 'lucide-react'
 import { Badge } from '@/components/ui/badge'
+import { Button } from '@/components/ui/button'
 import { cn } from '@/lib/utils'
 import type { ComponentThreat } from '../../types/threat-analysis'
 import { deriveThreatStatus } from '../../types/threat-analysis'
@@ -43,6 +44,7 @@ export function ComponentTreeItem({
   onSelectComponent,
   onToggleCollapsed,
   resolveTechName,
+  onRequestDeleteComponent,
 }: {
   treeNode: ComponentTreeNode
   componentThreats: ComponentThreat[]
@@ -51,6 +53,7 @@ export function ComponentTreeItem({
   onSelectComponent: (id: string) => void
   onToggleCollapsed: (id: string) => void
   resolveTechName: (value: string | undefined) => string
+  onRequestDeleteComponent: (component: { id: number; name: string }) => void
 }) {
   const { node, children, depth } = treeNode
   const Icon = nodeTypeIcons[node.type as string] || Cog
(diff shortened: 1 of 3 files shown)


vikram-s-narayan:
@BhavyaB19 - this is really, really cool work! Love how the delete icons only show up on hover. Also, deletion of threats is only possible on dismissed countermeasures which is the behavior we wanted because we we don't want a threat modeling team member to directly delete a threat that is part of a library. 

I'll test soon and merge or get back with a response.

vikram-s-narayan:
@sidd190 - would be good for you to incorporate this code into what you're working on with regard to https://github.com/precogly/precogly/issues/2

vikram-s-narayan:
@BhavyaB19  - One immediate design level issue that I just discovered is this: 

Currently, when a threat is "Dismissed", it shows up under "Dismissed Threats" in the Compliance Report. Please see screenshots below:

Dismissed Threats:
Screenshot 2026-05-16 at 4 47 58 AM


Show up in Compliance Section of Reports:
Screenshot 2026-05-16 at 4 49 33 AM

This is by _design_. Many compliance standards require that threats were considered and correctly handled. Now, if we completely delete a threat from our database, we're not able to show that the threat was considered in the first place. 

Moreover, deleting a threat also cascade-deletes its countermeasures, so it's not just the threat record that disappears ... but also the evidence of evaluated controls goes away.

Now, with this PR, we're introducing a potential source of confusion. A user could delete a threat and destroy the audit trail.

**Proposal**

Let us allow for the deletion of components as well as countermeasures. But not for the deletion of threats. Threats can only be dismissed but not deleted.

--
  **Code fix:** 

In `ComponentView.tsx`, in the dismissed threats section, remove the delete button entirely by dropping the condition that renders the `Trash2` icon. The block to remove:

  ```tsx
  {ct.backendThreatId !== undefined && (
    <Button
      variant="ghost"
      size="icon"
      className="h-8 w-8 opacity-0 group-hover:opacity-100 transition-opacity text-muted-foreground hover:text-destructive"
      onClick={() => {
        if (!ct.threatName || ct.backendThreatId === undefined) return
        setDeleteThreatConfirmFor({
          id: ct.id,
          name: ct.threatName,
          backendId: ct.backendThreatId,
          type: ct.threatType === 'dataflow' ? 'flow' : 'component',
        })
      }}
      aria-label={`Delete ${ct.threatName}`}
      title="Delete threat"
    >
      <Trash2 className="h-4 w-4" />
    </Button>
  )}

This sits inside the dismissed threats .map() callback. Remove it entirely so dismissed threats only show the "Restore" button.

sidd190:
> @sidd190 - would be good for you to incorporate this code into what you're working on with regard to #2

Hmmmm, coool! I've taken note of this for the integration!

BhavyaB19:
Makes sense, especially with the Compliance Report implications. I have removed the threat delete button as requested and also cleaned up the related unused threat deletion logic (mutation hooks, dialog handling, and associated function calls) since threats are no longer intended to be deletable.

PR #106, opened 2026-06-03
fix: New elements appear in the middle of the screen

### LINKED ISSUE #80: Clicking on one of the components on the toolbox should place the element in the middle of the current screen
When I click on one of the components on the top bar, new components are dropped on the right bottom of the model.
If I scroll to another location of the model, component is still dropped relatively to the whole model.

It would be better if new elements are dropped in the middle of the screen. Draw.io has a similar behavior - it's easier to see the component and use it.

aprv10:
## Summary
- Places toolbar added DFD nodes at the center of the currently visible canvas
- Uses React Flow screen to flow coordinate conversion so pan/zoom are respected

https://github.com/user-attachments/assets/e5d42091-db1d-447f-9748-6cfc714b363b

Fixes #80

### DIFF
--- a/frontend/src/features/dfd-editor/DFDEditor.tsx
+++ b/frontend/src/features/dfd-editor/DFDEditor.tsx
@@ -9,6 +9,7 @@ import {
   useReactFlow,
   useViewport,
   type Connection,
+  type XYPosition,
   addEdge,
 } from '@xyflow/react'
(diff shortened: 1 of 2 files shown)


aprv10:
@vikram-s-narayan can you review it and let me know if it needs any changing?

vikram-s-narayan (approved):
Looks good to me 👍

PR #108, opened 2026-06-06
feat: Track trust boundary crossings on data flows

### LINKED ISSUE #105: Data Flows Should Pick Up Trust Boundary Crossings Automatically
At present, when a data flow crosses a trust boundary, it does not contain any semantic meaning (i.e. which trust boundary am I crossing). Nor is there a way to asssociate which data flow crosses which trust boundary. There is a section to attach a trust zone crossing with a data flow but nothing to associate a trust boundary crossing. 

We need to create an association to track this information and also automatically detect boundary crossings by data flow arrows.

Image

This will allow us to build automated rule engines that trigger on boundary crossings.

sidd190:
Fixes #105 

[Screencast from 2026-06-07 00-06-54.webm](https://github.com/user-attachments/assets/be922672-29a4-468a-921c-74bd3be30b3b)

 What was built:
  1. Data model - DataFlow.trust_boundary_ids stores which trust boundaries a flow crosses
  2. Auto-detection - On save, the sync service analyzes source/target component zones and detects which trust boundaries are crossed using zone hierarchy traversal
  3. Cascade cleanup - When a trust boundary is deleted, a post_delete signal automatically removes that boundary ID from all data flows that reference it
  4. UI display -  EdgeEditPanel shows "Crosses Boundaries" with the boundary's label and security controls
  
 
 After quite a lot of debugging and sitting on the same thing, I feel like I'm missing something toh I request a review around the numbering of boundaries part for this PR, I seem to be skipping over the same thing I don't know. Though it might be working just fine now, so like just a careful review related to that.


vikram-s-narayan:
Cool stuff @sidd190 🚀

I'm hoping to test and review this in detail later today.

vikram-s-narayan:
Hi @sidd190 - sorry for the delayed review. 

I just discovered a bug. To reproduce: 
Create a trust zone. Then create two nested trust zones inside the parent trust zone. Then create a boundary between those two child trust zones. Then create a data flow between two components inside those two child trust zones. You'll see that the boundary crossing is not recognized. This can be seen from the attached image. In the set up on the left, you don't see a boundary crossing recognized. But in the set up on the right, you'll see that the boundary crossing is recognized.

Screenshot 2026-06-09 at 11 28 54 AM

vikram-s-narayan:
The following were flagged by Claude Code as "Dead code" and you can (after verification) remove them:
1. Three unused backend functions — _get_zone_ancestor_ids, _detect_boundary_crossings, and _get_all_boundaries_for_zones are defined but never called. They appear to be earlier single-flow versions that were replaced by the batch approach.                
2. Wasted loop in _sync_edges_to_dataflows — lines 280-291 call _get_all_boundaries_for_zones per dataflow edge, then the result is immediately overwritten by TrustBoundary.objects.all(). The double fetch of that table (lines 295-298) could also be collapsed into one query.
3. Dead boundary numbering code on the frontend — nextBoundaryIndex state, its ref, the useEffect syncing it, and boundaryDisplayId on edge data are all written but never read. The EdgeEditPanel computes the display number independently from the array index.
4. Dead fallback lookups in EdgeEditPanel — the snake_case and edge ID pattern lookups never execute because the camelCase middleware ensures the first lookup always matches.

vikram-s-narayan:
Finally, as a separate but related issue, it would be good to remove `crosses_trust_zone` — the boolean field on the DataFlow table that has never been set to True by any code path and is now redundant given trust_boundary_ids. It can be removed along with its references in the serializer, admin, views filter, and report service. But please verify before you remove :)

vikram-s-narayan:
@sidd190 - closing this PR as the codebase has moved forward since we had this discussion. We will need to come back to fixing the boundary crossing issue again in the future as the TM-BOM also has boundary crossing fields in it.

Issue #99, opened 2026-05-25
Full Taxonomy Packs: CWE, CAPEC, MITRE ATT&CK, MITRE ATLAS


vikram-s-narayan:
## Problem

The platform has "mini" taxonomy packs (`mini-cwe`, `mini-capec`, `mini-attack`) with a small curated subset of entries. These are sufficient for the demo `aws-mini` pack but not for production threat-library packs like `aws` and `aws-ai`, which need to reference a much broader set of entries.

MITRE ATLAS has no pack at all — it's needed for the upcoming `aws-ai` pack's AI/ML threat taxonomy.

## What Needs to Be Created

Four new taxonomy packs. They have no dependencies on each other and can be authored in any order:

```
libraries/packs/taxonomies/
├── cwe/                  # replaces mini-cwe
├── capec/                # replaces mini-capec
├── mitre-attack/         # replaces mini-attack
└── mitre-atlas/          # new
```

All four follow the same two-file structure (`pack.yaml` + `taxonomy.yaml`) used by the existing packs.

---

## 1. CWE (Common Weakness Enumeration)

**Location:** `taxonomies/cwe/`

**Source:** [CWE List (MITRE)](https://cwe.mitre.org/data/index.html)

**Scope:** The full CWE list has 900+ entries. Not all are relevant to threat modeling. Focus on:
- [CWE Top 25 Most Dangerous Software Weaknesses](https://cwe.mitre.org/top25/) (already partially covered by mini-cwe)
- CWEs referenced by [AWS FSBP](https://docs.aws.amazon.com/securityhub/latest/userguide/fsbp-standard.html) and [Prowler](https://docs.prowler.com/projects/prowler-open-source/en/latest/)
- CWEs commonly mapped in OWASP ASVS and OWASP Top 10
- Cloud/infrastructure-relevant CWEs (misconfiguration, access control, cryptography, injection)

**Estimated entries:** 80–120

**Data available for each entry:** MITRE provides structured data. Each entry needs:

```yaml
- external_id: CWE-79
  title: Improper Neutralization of Input During Web Page Generation (XSS)
  description: "The application does not properly neutralize user-controllable input before placing it in output used as a web page."
  reference_url: "https://cwe.mitre.org/data/definitions/79.html"
```

[...]

PR #143, opened 2026-07-03
feat: add mitre-atlas taxonomy pack

vishnushri19:
Refs #99
Adds a MITRE ATLAS taxonomy pack, this is the AI/ML threat taxonomy the aws-ai pack needs and currently doesn't have (there was no ATLAS pack at all before this).
Source: ATLAS release v2026.06 (current release, pulled from dist/v6/ATLAS-2026.06.yaml, not the STIX bundle,  ATLAS ships versioned snapshots, not an append-only object store like ATT&CK).
Scope: Full catalog, no curation. The issue estimated ~60-80 techniques, but the current release has grown since that estimate was written, ended up with 173 entries (103 top-level + 70 sub-techniques) across all 16 current ATLAS tactics. Checked for deprecated/revoked entries to filter out but there weren't any,  each ATLAS release is a curated "current state" snapshot so there's nothing historical sitting in it.
Schema note: taxonomy.yaml is flat (external_id, title, description, reference_url no parent/hierarchy field, confirmed against libraries/README.md and grepped the whole repo for any precedent). Sub-techniques are represented through the dotted ID convention (e.g. AML.T0008.004 under AML.T0008) and parent-prefixed titles (e.g. "Acquire Infrastructure: Serverless"), same pattern ATT&CK-style packs will need too.

Verification:
173/173 unique external_ids, zero dupes (checked independently, not just trusting the extraction script)
Random sample of reference_urls checked manually in browser, all resolve to real ATLAS technique pages
Imported locally, all 173 entries render clean in the taxonomy preview, zero errors/warnings from validate_pack()

Following the "one pack per PR" convention from the docs cwe, capec, and mitre-attack are up as separate PRs.

### DIFF
--- a/libraries/packs/taxonomies/mitre-atlas/pack.yaml
+++ b/libraries/packs/taxonomies/mitre-atlas/pack.yaml
@@ -1,0 +1,14 @@
+pack:
+  schema_version: 1
+  slug: mitre-atlas
+  name: MITRE ATLAS
+  description: Full catalog of MITRE ATLAS (Adversarial Threat Landscape for AI Systems)
+    tactics and techniques covering adversary behaviors that target AI/ML systems.
+  version: 1.0.0
+  pack_type: taxonomy
+  author: Precogly
+  tags:
+  - mitre-atlas
+  - adversarial-ml
+  - ai-security
+  - mitre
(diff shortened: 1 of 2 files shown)


vikram-s-narayan:
Amazing work!

PR #154, opened 2026-07-08
fix: api/schema stale field error and drift in docs

### LINKED ISSUE #153: [Bug] : Failing to load the api definition on localhost:8000/api/docs
Image

This seems like a legit bug for the api docs, I have applied all migrations but it seems like there's a problematic model field still in use at api/schema.

Image

Is this a local me-only problem or happens on everyone's side? Fix is fairly simple, just going through api/schema and finding the problematic fields should fix this if it's a general problem.

sidd190:
one backend api bug plus one tiny docs cleanup.

Fixes #153 

[Screencast from 2026-07-08 17-00-55.webm](https://github.com/user-attachments/assets/f3c57513-da4b-4614-955c-fbed90cd7c1a)

  1. The backend/API bug
      - File: backend/apps/threats/serializers.py:120
      - Removes stale organization from CountermeasureLibrarySerializer.
      - Fixes /api/schema/ crashing with:
        Field name organization is not valid for model CountermeasureLibrary

      - This also restores schema-backed API docs behavior.

  2. Small docs correction
      - File: docs/api/overview.md:165
      - Replaces nonexistent documented endpoints:
        /add_framework/, /remove_framework/
      - With real endpoints:
        /add_pack/, /remove_pack/

### DIFF
--- a/backend/apps/threats/serializers.py
+++ b/backend/apps/threats/serializers.py
@@ -120,7 +120,6 @@ class CountermeasureLibrarySerializer(serializers.ModelSerializer):
         model = CountermeasureLibrary
         fields = [
             "id",
-            "organization",
             "name",
             "description",
             "control_type",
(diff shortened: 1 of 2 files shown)


vikram-s-narayan (approved):
Looks great!

PR #159, opened 2026-07-11
fix: remove stale NOT NULL columns from packs tables that block pack seeding/import

### LINKED ISSUE #158: [Bug] Migration↔schema divergence: tier NOT NULL breaks pack seeding & import
Image

Image

The live `packs_librarypack` table has `tier` (and `source`, `content`, `industries`, `install_count`, `is_published`, …) as `NOT NULL` with no default, but the current model and `0001_initial.py` no longer define them. `update_or_create` INSERTs without `tier` → `null value in column "tier" ... violates not-null constraint`. Existing rows only get UPDATEs and survive; **new** packs (capec, cwe, mitre-attack, owasp-top-10, owasp-asvs) hit INSERT and **all fail to seed/import**. 

Root cause: a "simplify pack directory structure" commit deleted those fields and rewrote `0001_initial` in place, with no `RemoveField` migration; `django_migrations` still records a now-deleted `packs.0002_add_standard_requirement_mapping`. Recorded history has diverged from the code

The seed also has 5 failing imports hence only having 10 total pre imported packs. It also causes import failure on the library packs. Attached screenshots for both of them.

sidd190:
## Summary

- Adds migration to drop 12 stale NOT NULL columns from `packs_librarypack` and `packs_librarypackdependency` that were removed from the models but never had a corresponding migration
- Cleans phantom `0002_add_standard_requirement_mapping` migration record from `django_migrations`
- Safe for both existing databases and fresh installs (checks column existence before dropping)

Fixes #158 

## Problem

A prior commit simplified the `LibraryPack` model by removing fields (`tier`, `source`, `content`, `industries`, `repository_url`, `documentation_url`, `icon_url`, `install_count`, `is_published`, `published_at`) and rewrote `0001_initial.py` in place — but never generated a `RemoveField` migration. The same happened for `version_constraint` and `is_optional` on `LibraryPackDependency`.

Existing databases still have these as NOT NULL columns with no default, so any INSERT for a new pack fails:

Skipped: taxonomies/capec — Import failed: null value in column "tier" of relation
"packs_librarypack" violates not-null constraint

This causes 5 of 10 seeded packs to fail (CAPEC, CWE, MITRE ATT&CK, OWASP Top 10, OWASP ASVS) and `aws-mini` to fail on the dependency table. Users see only 10 packs in the Imported tab instead of 15.

[Screencast from 2026-07-11 19-19-41.webm](https://github.com/user-attachments/assets/ce01690c-4203-4394-be89-5c62d298a626)

### DIFF
--- a/backend/apps/packs/migrations/0002_remove_stale_librarypack_columns.py
+++ b/backend/apps/packs/migrations/0002_remove_stale_librarypack_columns.py
@@ -1,0 +1,66 @@
+"""
+Remove stale columns from packs_librarypack and packs_librarypackdependency.
+
+These columns were removed from the models but never had a corresponding migration.
+Existing databases still have these NOT NULL columns, causing INSERT failures for
+any new pack or dependency.
+"""
+
+from django.db import migrations
+
+
+STALE_COLUMNS = {
+    "packs_librarypack": [
+        "tier", "source", "content", "industries",
+        "repository_url", "documentation_url", "icon_url",
+        "install_count", "is_published", "published_at",
+    ],
+    "packs_librarypackdependency": [
+        "version_constraint", "is_optional",
+    ],
+}
+
+
+def remove_columns_if_exist(apps, schema_editor):
+    """Drop stale columns only if they exist (safe for fresh installs)."""
+    connection = schema_editor.connection
+    for table, stale_columns in STALE_COLUMNS.items():
+        with connection.cursor() as cursor:
+            cursor.execute(
+                "SELECT column_name FROM information_schema.columns "
+                "WHERE table_name = %s", [table]
+            )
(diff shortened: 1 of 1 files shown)


vikram-s-narayan:
@sidd190 ... This only affects databases originally created under v0.1.0. Anyone on v0.2.0 or starting fresh never has these columns.

The fix uses raw SQL instead of Django migrations. The information_schema query, manual ALTER TABLE, and direct DELETE FROM django_migrations bypass Django's migration framework entirely. This works, but it's not idiomatic Django which is preferred.

To fix the problem you just need to do `docker compose down -v` && `docker compose up --build` to start fresh.

Happy to re-open the issue if needed :)

PR #161, opened 2026-07-11
Add a data NULL check for tight code quality and UX

### LINKED ISSUE #160: bug: pack preview endpoint returns HTTP 200 with empty body instead of 404 when pack slug doesn't match on-disk YAML
## Summary

The `preview` endpoint for imported packs (`GET /api/packs/{id}/preview/`) returns HTTP 200 with ab empty/null body instead of a proper 404 when the pack's database slug doesn't match any on-disk `pack.yaml` slug.

I found while investigating #158  (pack seeding failure). The old mini packs (`mini-capec`, `mini-asvs`, `mini-attack`, `mini-cwe`, `owasp`) had DB slugs that diverged from the current on-disk slugs (`capec`, `owasp-asvs`, `mitre-attack`, `cwe`, `owasp-top-10`)   Clicking Preview on any of these in the Imported tab showed a blank panel - the API returned 200 OK with nothing.

The data mismatch was resolved by the re-seed in the fix, but the code-level issue remains: any pack whose DB slug doesn't match an on-disk `pack.yaml` will silently return an empty 200 instead of a clear 404.

- It ended up with the frontend preview panel showing blank/loading state with no error feedback - the user would have no idea why the preview isn't working
- Clients cannot distinguish "pack not found on disk" from a successful return
- The sibling endpoint `preview_from_source` correctly returns 404 for missing packs, making the behavior inconsistent
- Can be triggered by any pack imported via API or CycloneDX whose slug dary

You can actually see the packs not loading the preview and reproduce this if #159 hasn't yet been merged and migrations haven't run by going to the imported packs and trying to preview them in the library or reverting the migration once back to 0001 only.

sidd190:
Fixes #160 

Simple fix, just added a NULL check and return in case there's a mismatch of endpoints. This had come up as an actual bug when testing around #158.

You can actually see the packs not loading the preview and reproduce this if #159 hasn't yet been merged and migrations haven't run by going to the imported packs and trying to preview them in the library or reverting the migration once back to 001 only.

### DIFF
--- a/backend/apps/packs/views.py
+++ b/backend/apps/packs/views.py
@@ -247,6 +247,11 @@ class LibraryPackViewSet(viewsets.ReadOnlyModelViewSet):
         """
         pack = self.get_object()
         preview_data = get_pack_preview_from_database(pack)
+        if preview_data is None:
+            return Response(
+                {"detail": "Pack source files not found on disk."},
+                status=status.HTTP_404_NOT_FOUND,
+            )
         return Response(preview_data)
 
     @action(detail=False, methods=["get"])


vikram-s-narayan:
@sidd190 - thanks for the fix - the null check itself is correct, but this can only trigger when the database has pack  slugs that don't match any on-disk YAML, which doesn't happen under normal operation. The scenario you hit was caused by the stale v0.1.0 data from #158, which is resolved by re-seeding. Since the only path to this state is the stale v0.1.0 data (which is resolved by re-seeding), I don't think we need to carry extra defensive code for it. Closing this
one out.

Happy to reopen if you would like to discuss this further.

Issue #209, opened 2026-07-17
[Security bug] : Security Team Permission Is Org-Agnostic


sidd190:
IsSecurityTeam currently checks only whether the user is security team in any organization:
 request.user.organization_memberships.filter(role="security_team").exists()

 It does not check whether the target object belongs to the same organization where the user has that role. OrganizationMemberViewSet and BusinessUnitViewSet use this permission class for writes. Their querysets return objects from organizations the user belongs to, but a user can belong to Org B as a normal member and Org A as security team. Because the permission check is global, security-team status in Org A grants write power over Org B objects.

  Effect:

  - Cross-org privilege escalation.
  - A user can edit business units or organization memberships in an org where they are only a regular member.
  - Tenant isolation is weakened.
  - Organization admins/security teams cannot rely on org-scoped roles.

Screen recording TBA

AlvinKuruvilla:
Was this fixed? At 3235fd6 it looks unfixed.

`IsSecurityTeam.has_permission` (`apps/core/permissions.py:19-21`) still asks only whether the user is security team *anywhere*:

```python
return request.user.organization_memberships.filter(
    role="security_team",
).exists()
```

The class defines no `has_object_permission`.

Both viewsets still pair that with a queryset covering every organization the user belongs to in any role:

- `OrganizationMemberViewSet` — `apps/organizations/views.py:130`
- `BusinessUnitViewSet` — `apps/organizations/views.py:157`

Both resolve scope with `org_ids = user.organization_memberships.values_list("organization_id", flat=True)` and no role filter, so the scenario described above still holds: security team in Org A plus ordinary membership in Org B yields write access to Org B's memberships and business units. `git log -- apps/core/permissions.py` shows no commit addressing this since the close.

Worth noting the fix pattern already exists twenty lines further down the same file. `IsTeamMember.has_object_permission` (`apps/organizations/views.py:184`) scopes correctly, against the object's own organization:

```python
if obj.organization.members.filter(
    user=request.user, role="security_team"
).exists():
    return True
```

Role handling for the MCP server depends on whether security-team status is org-scoped
cc @sidd190 @vikram-s-narayan

sidd190:
> Was this fixed? At [3235fd6](https://github.com/precogly/precogly/commit/3235fd6181acca5f5e8ad30dee0f64fb6c9b7022) it looks unfixed.

[...]

PR #228, opened 2026-07-19
fix/CycloneDX import errors and scope IsSecurityTeam permission

### LINKED ISSUE #224: CycloneDX import can persist invalid control statuses and fail on duplicate affected assets
The CycloneDX 2.0 TM-BOM import path has two robustness problems that can make imported threat models unreliable.

First, the CycloneDX status mapping maps decommissioned to the local string decommissioned. That value is not one of Precogly’s valid
InstanceCountermeasure.Status choices. Because Django model choices are not enforced by a database check constraint here, the invalid status can be persisted. Once stored, downstream threat status, residual scoring, reports, and UI filters may handle the countermeasure inconsistently.

(Note : This bug specifically for decommissioned was recently fixed and added to enum but there's still a need for validation guard in the enums, it's moer of a defensive robustness thing now that this was fixed, but still a good to have?)

  Second, abstract-threat import creates instance threats with .create() for every affectedAssets entry. ComponentInstanceThreat and DataFlowInstanceThreat have uniqueness constraints on (component, threat_library) and (data_flow, threat_library). If a CycloneDX BOM repeats the same affected asset for the same abstract threat, or otherwise resolves duplicate abstract threats to the same asset/library pair, import can raise an IntegrityError. Since the import runs as one transaction, that rolls back the whole threat model import.

  Impact

  - Imported threat models can contain invalid internal countermeasure statuses.
  - Reports and scoring may become misleading after import.
  - Reasonable external BOMs with duplicate affected-asset references can fail completely.
  - Users get an opaque import failure instead of a partial import or warning.
  - This weakens trust in CycloneDX/TM-BOM interoperability, which is a core import/export story.

Problem 1 

Image
Image

Problem 2 

Image
Image

Also something to note is the error handling here, currently just shows error codes, this can be paired in the fix later, when db validation or idempotency is added probably.

sidd190:
## Summary

Fixes #224 and #209

### CycloneDX Import Improvements

- CycloneDX import now validates mapped control statuses against `InstanceCountermeasure.Status.choices` before persisting. If a future CycloneDX status maps to an invalid local value, the import now defaults to `gap` and surfaces a warning instead of attempting to save an invalid choice to the database.
- CycloneDX abstract-threat import now uses `.get_or_create()` instead of `.create()` for `ComponentInstanceThreat` and `DataFlowInstanceThreat`. Previously, if a BOM listed the same asset multiple times in `affectedAssets`, the import would raise an `IntegrityError` due to the `(component, threat_library)` unique constraint, causing the entire import to fail with an opaque HTTP 400 response. Duplicate entries are now safely deduplicated, and a user-facing warning is included in the import summary (displayed in the existing yellow warning banner in the import dialog).

### Permission Fix

- `IsSecurityTeam` previously implemented only `has_permission`, which checked whether a user had the `security_team` role in *any* organization. As a result, a user who was `security_team` in **Organization A** but only a member in **Organization B** could still modify Organization B's business units, member records, and other organization-scoped resources.
- Added `has_object_permission`, which resolves the owning organization of the target object and verifies the user's role within that specific organization before allowing writes.
- Extracted `_get_organization` into a shared module-level helper so both `IsSecurityTeam` and `CanWrite` use the same organization resolution logic.
- Global catalog resources (where no owning organization can be resolved) continue to follow the existing behavior and are handled correctly.

Screenshot From 2026-07-19 22-41-19

### DIFF
--- a/backend/apps/core/permissions.py
+++ b/backend/apps/core/permissions.py
@@ -5,10 +5,50 @@ Core permission classes for RBAC.
 from rest_framework import permissions
 
 
+def _get_organization(obj):
+    """Walk FK paths to find the organization for an object."""
+    # Direct organization FK
+    if hasattr(obj, "organization_id") and obj.organization_id:
+        return obj.organization
+
+    # threat_model -> organization
+    if hasattr(obj, "threat_model_id") and obj.threat_model_id:
+        return obj.threat_model.organization
+
+    # component -> orgsystem -> organization
+    if hasattr(obj, "component_id"):
+        component = obj.component
+        if component and hasattr(component, "orgsystem") and component.orgsystem:
+            return component.orgsystem.organization
+
+    # data_flow -> source_component -> orgsystem -> organization
+    if hasattr(obj, "source_component_id"):
+        source = obj.source_component
+        if source and source.orgsystem:
+            return source.orgsystem.organization
+
+    # orgsystem -> organization
+    if hasattr(obj, "orgsystem_id") and obj.orgsystem_id:
+        return obj.orgsystem.organization
+
+    # orgsystem (the object itself IS an Orgsystem)
+    if hasattr(obj, "organization_id") and obj.__class__.__name__ == "Orgsystem":
+        return obj.organization
+
+    return None
+
+
 class IsSecurityTeam(permissions.BasePermission):
     """
     Restricts write operations to security team members.
     Read operations are allowed for all authenticated users.
+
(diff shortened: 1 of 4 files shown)


vikram-s-narayan (approved):
Hi @sidd190 - this is awesome work. The only thing I'd change for the future is to split these two fixes into two separate PRs  ... but otherwise, this is fantastic. Thank you!

PR #233, opened 2026-07-19
Fix status/effectiveness mapping for in_progress and implemented coun…

### LINKED ISSUE #207: Bug : In-Progress/Implemented Controls Mark Threats Mitigated But Give 0 Credit
The app has two separate interpretations of countermeasure status, and they disagree.

In backend/apps/threats/services.py, threat status derivation treats anything that is not clearly gap, planned, or waived as effectively mitigated. That means a countermeasure marked in_progress or implemented can cause the related threat to show as mitigated.

But residual score calculation uses a separate effectiveness map. That map does not include in_progress or implemented, so both statuses fall back to 0.0 effectiveness. The result is contradictory: the UI/report can say the threat is mitigated, while the residual risk calculation behaves as if the countermeasure did nothing.

  Effect:

  - Users may believe a threat is mitigated when risk scoring still treats it as untreated.
  - Reports can present inconsistent security posture.
  - Compliance/risk review becomes hard to trust because status and score disagree.
  - This is especially risky for implemented, which sounds completed but receives no scoring credit.

Screen Recording to be added in a bit

sidd190:
## Summary

- `in_progress` and `implemented` countermeasures were missing from `STATUS_EFFECTIVENESS_FALLBACK`, defaulting to `0.0` — meaning they provided **zero risk reduction** in residual score calculations despite marking threats as "mitigated"
- `recalculate_threat_status()` treated `in_progress` and `decommissioned` as fully mitigated (same bucket as `verified`/`platform`), which contradicts their partial/zero effectiveness
- Added missing statuses to the fallback map: `in_progress: 0.3`, `implemented: 0.7`, `decommissioned: 0.0`
- Fixed status derivation: `in_progress` and `decommissioned` now correctly produce `addressable`, not `mitigated`

## What changed

| Status | Before (effectiveness → threat status) | After |
|--------|---------------------------------------|-------|
| `in_progress` | 0.0 → mitigated | 0.3 → addressable |
| `implemented` | 0.0 → mitigated | 0.7 → mitigated |
| `decommissioned` | 0.0 → mitigated | 0.0 → addressable |

Fixes #207

### DIFF
--- a/backend/apps/threats/services.py
+++ b/backend/apps/threats/services.py
@@ -13,9 +13,12 @@ from .scoring.registry import get_scoring_methods, score_to_level
 STATUS_EFFECTIVENESS_FALLBACK = {
     "verified": 1.0,
     "platform": 1.0,
+    "implemented": 0.7,
+    "in_progress": 0.3,
     "planned": 0.5,
     "gap": 0.0,
     "waived": 0.0,
+    "decommissioned": 0.0,
 }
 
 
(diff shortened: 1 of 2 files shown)


vikram-s-narayan (approved):
nicely done!

PR #250, opened 2026-07-28
fix: add rate limiting to login, registration, and password reset endpoints

### LINKED ISSUE #249: No rate limiting or account lockout on authentication endpoints
The login (/api/auth/login/), registration (/api/auth/registration/), and password reset (/api/auth/password/reset/) endpoints have no rate limiting, throttling, or account lockout mechanism. An attacker can make unlimited authentication attempts without any restriction.

Reproduction:

### 50 rapid failed login attempts - all return HTTP 400, no lockout
for i in $(seq 1 50); do
  curl -s -o /dev/null -w "Attempt $i: HTTP %{http_code}\n" \
    http://localhost:8000/api/auth/login/ -X POST \
    -H "Content-Type: application/json" \
    -d "{\"email\":\"admin@precogly.dev\",\"password\":\"wrong$i\"}"
done

### Correct password still works after 50 failures
curl -s http://localhost:8000/api/auth/login/ -X POST \
  -H "Content-Type: application/json" \
  -d '{"email":"admin@precogly.dev","password":"admin123"}'

Rate limiting is essential and has more than a few direct impacts that cause security issues like brute force attacks/credential stuffing, spam accounts creation, password reset endpoint can be used for finding random user emails. Even though some might be multi tenancy problems in general, but even on self hosting/sandboxing, this is a problem per instance if left as is.

sidd190:
## Summary

- Add per-IP rate limiting to authentication endpoints to prevent brute-force and credential stuffing attacks
- Login: 5 requests/min, Registration: 3 requests/min, Password reset: 3 requests/min
- Uses DRF's built-in throttling with scoped `AnonRateThrottle` subclasses, no new deps introduced.

Fixes #249 

## Reproduction

Before this fix, unlimited login attempts are possible with no lockout:

```bash
for i in $(seq 1 50); do
  curl -s -o /dev/null -w "Attempt $i: HTTP %{http_code}\n" \
    http://localhost:8000/api/auth/login/ -X POST \
    -H "Content-Type: application/json" \
    -d "{\"email\":\"admin@precogly.dev\",\"password\":\"wrong$i\"}"
done
# All 50 return HTTP 400 - no lockout, no rate limit

After this fix, the 6th attempt returns HTTP 429 with a Retry-After header.

Changes

- backend/apps/core/throttles.py - three scoped throttle classes (login, registration, password_reset)
- backend/apps/core/auth_views.py - throttled subclasses of dj-rest-auth LoginView, RegisterView, PasswordResetView
- backend/config/urls.py - override auth URLs with throttled views (placed before dj-rest-auth includes)
- backend/config/settings/base.py - throttle rate configuration in REST_FRAMEWORK

### DIFF
--- a/backend/apps/core/auth_views.py
+++ b/backend/apps/core/auth_views.py
@@ -1,0 +1,20 @@
+from dj_rest_auth.registration.views import RegisterView
+from dj_rest_auth.views import LoginView, PasswordResetView
+
+from apps.core.throttles import (
+    LoginRateThrottle,
+    PasswordResetRateThrottle,
+    RegistrationRateThrottle,
+)
+
+
+class ThrottledLoginView(LoginView):
+    throttle_classes = [LoginRateThrottle]
+
+
+class ThrottledRegisterView(RegisterView):
+    throttle_classes = [RegistrationRateThrottle]
+
+
+class ThrottledPasswordResetView(PasswordResetView):
+    throttle_classes = [PasswordResetRateThrottle]
--- a/backend/apps/core/tests/test_auth_throttles.py
+++ b/backend/apps/core/tests/test_auth_throttles.py
@@ -1,0 +1,123 @@
+"""Tests for authentication rate limiting.
+
+Verifies that login, registration, and password reset endpoints enforce
+per-IP request limits to prevent credential stuffing, registration
+flooding, and password reset abuse.
+
+Uses the production throttle rates from settings (login: 5/min,
+registration: 3/min, password_reset: 3/min).
+"""
+
+from django.contrib.auth import get_user_model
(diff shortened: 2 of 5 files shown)


vikram-s-narayan:
Closing without merge for now. Precogly is designed for internal/on-prem deployment, so brute-force protection on auth endpoints is low priority at this stage (v0.3.0). More importantly, the current implementation uses per-IP throttling via AnonRateThrottle, which means orgs running behind a reverse proxy (most deployments) would have all users sharing one IP address. A few login  attempts from different people would lock out the entire organization. We can revisit this later with proxy-aware IP resolution or an opt-in configuration flag.

Issue #219, opened 2026-07-19
Unify the Add Threat modal and retire the standalone AI affordance


AlvinKuruvilla:
**Context:** Adding a threat is one action that branches three ways — from the library, custom, or AI-suggested — but AI currently lives as a separate owl button beside the add flow.

**Done looks like:** One Add Threat entry point offering all three paths, and the standalone owl affordance removed from both the threats panel and the DFD editor.

**Open questions:** Whether the AI tab should offer pulling from the model's own knowledge in addition to ranking library pack threats — more coverage, but it burns tokens and can hallucinate, so it would need to be a clearly labelled separate choice.

From the [2026-07-18 sync](https://github.com/precogly/precogly/discussions/214).

AlvinKuruvilla:
**AI isn't a third source, it's a sort order**

I went looking at how `suggest.py` actually works before building the third tab, and it doesn't do what the description assumes. It gets the same pack threats the library picker lists, asks the model to pick the relevant ones and say why, then throws away anything whose id wasn't in that list. It can't produce a threat the library path couldn't.

So this isn't three branches:

```text
what the description assumes          what the code does

  Add Threat                            Add Threat
  ├── From library                      ├── From library ──┐
  ├── Custom                            │                  │  rank + rationale
  └── AI suggested                      │   [AI] ──────────┘  (same pool, reordered)
                                        └── Custom
```

Two sources. AI is a lens on one of them. A third tab would make you choose between browsing the library and having it ranked, which isn't a choice — it's the same list.

#167 is what makes this worth doing. Now that pack threats don't auto-populate, you open the picker to up to 40 candidates in no order at all. Picking through those is the actual chore, and it's the exact thing `suggest.py` answers: here are the ones that matter for this component, here's why, here's a severity. Triage, not generation.

**Revised done-looks-like:** one Add Threat entry point, two paths. The library list can be ranked, each row showing why it applies and a suggested severity, accepted one at a time through the normal create path. Standalone owl gone from the threats panel and the DFD editor.

**Where the owl goes.** Inside the modal, on the ranking control:

```text
Add Threat
┌──────────────────────────────────────────────┐
│  From library          Custom                │
├──────────────────────────────────────────────┤
│  [ search…]            (🦉) Rank by relevance│
│                                              │
│  ☐ Spoofing of API Gateway identity          │
│  ☐ Unvalidated request forwarding            │
│  ☐ ...                                       │
└──────────────────────────────────────────────┘
```

[...]

PR #256, opened 2026-07-28
fix(threats): exclude flow-scoped threats from the component library picker

AlvinKuruvilla:
## The problem

A component library maps two different things through the same join table: the threats a component carries itself, and the threats its *connections* carry. An API gateway maps eavesdropping and replay so that flows touching it inherit them — those threats belong on the data flow, never on the gateway.

The `component_id` filter on `/threat-library/` narrowed to the component's library but ignored `applies_to`, so the Add Threat picker offered flow-scoped threats when adding to a component. Users added them there, producing component threats that the pack never intended to exist.

## The change

Adds the `applies_to in (COMPONENT, BOTH)` filter to that queryset, mirroring `apps.threats.ai.suggest.candidate_library_threats`, which has always filtered correctly. The AI ranker therefore never had this bug — the two paths express the same conceptual scope and now apply the same rule.

The filter is deliberately scoped to the `component_id` branch only. Without `component_id` the endpoint is the whole catalogue, where a flow threat is a real entry that must still be listed, or browsing a pack would silently hide part of it. The third test pins that.

## Verification

```
$ docker compose exec backend python -m pytest apps/threats/tests/test_threat_library_scoping.py -q
3 passed in 5.19s
```

The fixture is the exact shape that produced the bug in the `aws-mini` pack: one API Gateway library mapping one threat of each scope.

## Not in this PR

The Add Threat dialog rework (#219) touches the same picker from the frontend side and ships separately. This fix stands on its own and does not depend on it.

### DIFF
--- a/backend/apps/threats/tests/test_threat_library_scoping.py
+++ b/backend/apps/threats/tests/test_threat_library_scoping.py
@@ -1,0 +1,94 @@
+"""What the ``component_id`` filter on ``/threat-library/`` may offer.
+
+A component library maps two different things through the same join table: the
+threats a component carries itself, and the threats its *connections* carry. An
+API gateway maps eavesdropping and replay so that flows touching it inherit
+them — those threats belong on the data flow, never on the gateway.
+
+The picker filtered by component library but not by ``applies_to``, so it
+offered flow-scoped threats when adding to a component, and users added them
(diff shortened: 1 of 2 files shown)


vikram-s-narayan (approved):
awesome catch and fix!

PR #257, opened 2026-07-28
feat(dfd-editor): fold AI ranking into Add Threat and retire the owl

### LINKED ISSUE #219: Unify the Add Threat modal and retire the standalone AI affordance

AlvinKuruvilla:
Closes #219.

## Demo

https://github.com/user-attachments/assets/3328b281-1bd1-4782-8c07-c72df43c15e2

## What changes

Per the 2026-07-18 sync, Add Threat becomes the single entry point with three branches — library, custom, and AI-ranked. The separate suggest-threats owl in the component header goes away. Two controls that both added threats, one of which was easy to miss entirely, become one.

`SuggestThreatsOwl.tsx` is deleted (−340). Its toggle, its error handling and its route to provider settings were all things any AI affordance needs, so the first commit lifts them out as `OwlToggle` and `AiErrorState` rather than duplicating them inside the dialog. `OwlMark` itself stays as the AI glyph.

Read it as two commits — the extracted primitives, then the dialog.

## Decisions worth a reviewer's attention

**Ranking is a third state on the existing scope toggle, not a fourth cell in a matrix.** `candidate_library_threats` only ever ranks the component's own pool, so ranking and "show all" are mutually exclusive by construction. Turning "show all" on switches ranking off rather than leaving it stale over a list it did not rank, and the Rank control goes inert with a tooltip saying why.

**Ranking is component-only.** The suggest endpoint takes a component id and has no dataflow equivalent, so dataflows never see the control.

**One LLM call per open.** Toggling Rank off and back on reuses the ranking already fetched rather than re-billing the organization.

**`resetForm` now runs on open, and this is the part I'd look at hardest.** Neither call site unmounts the dialog — both re-render it in place with a new `targetId` — so before this change every field survived a Cancel and reappeared against whatever component was opened next. For the ranking that was not merely untidy: suggestions stayed on screen labelled "Ranked for this component", the `!suggest.data` guard suppressed the refetch that would have corrected them, and accepting one filed another component's threat.

I reproduced this against two components with disjoint candidate sets before fixing it: rank the API Gateway, Cancel, open Lambda, and Lambda showed the API Gateway's three suggestions with no request fired.

The alternative was `key={targetId}` at the call sites, which preserves the cached ranking across reopens of the same component and so avoids a second metered call. I went the other way: it fixes only the cross-component leak, leaves the same-component field staleness (a half-typed custom threat name would still carry over), and lives in two call sites a third will forget. Resetting on open rather than on close keeps the fields populated through the closing animation instead of blanking them mid-fade.

## Verification

```
$ docker compose exec frontend npx tsc --noEmit -p tsconfig.app.json   # exit 0
$ docker compose exec frontend npx eslint <the four changed files>     # clean
```

Manually, against LM Studio (`openai/gpt-oss-20b`) on the seeded Sample Threat Model. The seed attaches every applicable threat to each component and `candidate_library_threats` excludes them all, so `suggest` returns `[]` everywhere until rows are deleted to open a gap — worth knowing for anyone reproducing this. With gaps opened on the API Gateway and Lambda:

- ranks, with rationale and pack attribution per row; already-attached threats correctly absent
- toggling Rank off/on fires no second POST
- "show all" switches ranking off, clears the selection, and blocks the control
- dataflows show no Rank control and fire no availability request
- model stopped → the 503 state, with a working retry
- saturated component → "nothing left to suggest" rather than the no-library-link message
- the cross-component leak above, before and after the fix

No automated frontend tests: the repo has no test infrastructure on that side (no Vitest, no Testing Library, no Playwright). Standing that up is its own decision and its own PR, not something to smuggle in here — but the state machine in this dialog is exactly the kind of thing that wants pinning, and I would rather it were.

## Deliberately not in this PR

- **`useFlowThreats` reads only page 1.** `DataFlowInstanceThreatViewSet` paginates at `PAGE_SIZE: 20` with no `page_size_query_param`, so a flow carrying more than 20 threats gets an incomplete already-added set and is re-offered threats the backend will reject. Fixing it properly needs a backend change — either `pagination_class = None` on that viewset, which pushes against e7d057d ("Fix unbounded catalog endpoints"), or a page-size param — and that is not a call to make inside a UI PR.
- **The `applies_to` scoping fix** for the same picker is #256, independent of this.


vikram-s-narayan (approved):
thank you Alvin. I always learn so much from your code and the way you approach things!

PR #277, opened 2026-08-03
feat: AI-powered DFD generation from architecture diagrams 

vikram-s-narayan:
## Summary                                                                                                                        
                                                                                                                                    
  Upload an architecture diagram image and let AI generate a first-cut DFD on the canvas. Two-step flow:
                                                                                                                                    
  - **Step 1 (Vision):** Extracts components, data flows, trust zones, and asks clarifying questions                                
  - **Step 2 (Generation):** Produces canvas data with component library matching and algorithmic layout correction                 
                                                                                                                                    
Follows the same provider-resolution, metering, and trust-boundary patterns as `suggest_threats`. Nothing is persisted until the user accepts the generated diagram.

### DIFF
--- a/backend/apps/ai/providers/base.py
+++ b/backend/apps/ai/providers/base.py
@@ -17,6 +17,7 @@ key is plaintext and the only job left is to make the call.
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from decimal import Decimal
+from typing import Any
 
 
 class AIProviderError(Exception):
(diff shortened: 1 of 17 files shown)


vikram-s-narayan:
This is a work in progress. Adding this here so @AlvinKuruvilla can take a look and give feedback. @AlvinKuruvilla - I went ahead and built out this feature as we needed it for the BlackHat demo. We can improve this over time.

AlvinKuruvilla:
Awesome, thanks @vikram-s-narayan. Just for provenance's sake,  noting that this is related to #288, #293, #294; however, these are an extension to this flow where we can use a file like a TM-BOM as a reference, with or without an image (not sure about that yet)

PR #301, opened 2026-08-11
fix: image export cropping and add an export options dialog

### LINKED ISSUE #280: Export image button improvements
When we use the Export Image button, a copy of a diagram is directy downloaded.
Downloaded image has lots of blank area around the diagram. Same image which is not sized correctly is also used in the docx report. 

1. Default image export should be cropped only to the area where components actually exist - reducing empty space around it.
2. A popup form asking for more options when Export button is clicked would be useful. It can offer following extra options:
   a) Use transparent background
   b) Export only selected elements

Possibility to export only selected elements is very useful because often diagrams might have elements that we don't necessery want to put into a report. 

Check draw.io image export feature as an example.

KhaledZer:
Closes #280

The exported diagram images had a ton of blank space around the actual diagram — turns out we were passing html-to-image a canvas twice the size of the content while styling the element at 1x, so the diagram only filled the top-left quarter. The same broken image ended up in the Word report since both go through the same helper, so fixing the capture fixed both.

While at it, I replaced the PNG/SVG dropdown with a small export dialog (similar to draw.io's):

- pick PNG or SVG
- transparent background — default is now solid white; the old exports were accidentally transparent, which is part of why they looked so empty in viewers
- export only selected elements — useful when a diagram has parts you don't want in a report. Selected edges bring their endpoint nodes along so nothing dangles, and edges between exported nodes are included automatically. Disabled with a hint when nothing is selected.

A few smaller things that came out of this:

- the export now deselects everything before capturing, so selection borders and delete buttons no longer show up in the image (they did before), then reselects afterwards. This goes through React Flow's `select` changes so it doesn't pollute undo history or trigger autosave.
- exporting an empty canvas is blocked, and a failed export shows a toast instead of silently closing the dialog like it worked
- added `data-id` to the edge label / threat badge portals so excluded edges don't leave floating labels behind in selected-only exports

Both the normal editor and the guest editor share the same dialog. The export functions kept their signatures (just one optional options param added), and the Word report flow only picks up the crop fix — nothing else in reports was touched.

Tested: build + lint clean, plus manually checked cropping, transparency, selected-only export, selection restore, undo/dirty state, and the guest Word report embed.

### DIFF
--- a/frontend/src/features/dfd-editor/DFDEditor.tsx
+++ b/frontend/src/features/dfd-editor/DFDEditor.tsx
@@ -44,7 +44,7 @@ import { useBoundaryMode } from './hooks/useBoundaryMode'
 import type { DiagramNode, DiagramEdge, DataFlowEdge, TrustBoundaryEdge } from './types'
 import { useCreateNode, useHandleDrop } from './hooks/useCreateNode'
 import { type DFDNotationStyle, NOTATION_NODE_SIZES } from './types/notation'
-import { exportDiagramImage } from './lib/export-diagram-image'
+import { exportDiagramImage, type ExportImageOptions } from './lib/export-diagram-image'
 
 function DFDEditorContent() {
   const { diagramId, id: threatModelId } = useParams<{ id: string; diagramId: string }>()
@@ -430,13 +430,13 @@ function DFDEditorContent() {
 
   // Export diagram as image
   const handleExportImage = useCallback(
-    (format: 'png' | 'svg') => {
+    (format: 'png' | 'svg', options?: ExportImageOptions) => {
       if (!reactFlowWrapper.current) return
       const filename = (diagramTitle || 'diagram')
         .replace(/[^a-zA-Z0-9-_ ]/g, '')
         .replace(/\s+/g, '-')
         .toLowerCase()
-      exportDiagramImage(format, filename, reactFlowWrapper.current, nodes, getViewport, setViewport, getNodesBounds)
+      return exportDiagramImage(format, filename, reactFlowWrapper.current, nodes, getViewport, setViewport, getNodesBounds, options)
     },
     [diagramTitle, nodes, getViewport, setViewport]
   )
--- a/frontend/src/features/dfd-editor/components/DiagramToolbar.tsx
+++ b/frontend/src/features/dfd-editor/components/DiagramToolbar.tsx
@@ -1,14 +1,8 @@
-import { memo, useCallback } from 'react'
+import { memo, useCallback, useState } from 'react'
 import { useNavigate } from 'react-router-dom'
 import type { XYPosition } from '@xyflow/react'
 import { User, Server, Cog, Database, Shield, Box, ArrowUp, LayoutTemplate, ShieldAlert, ShieldCheck, ImageDown } from 'lucide-react'
 import { Button } from '@/components/ui/button'
-import {
-  DropdownMenu,
-  DropdownMenuContent,
-  DropdownMenuItem,
-  DropdownMenuTrigger,
-} from '@/components/ui/dropdown-menu'
 import { Separator } from '@/components/ui/separator'
 import {
   Select,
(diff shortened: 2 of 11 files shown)


KhaledZer:
image



# How the export image fix + dialog was implemented

## The bug

`exportDiagramImage` handed html-to-image a canvas sized at **2x** the diagram bounds (for retina-quality output) but then styled the captured element back to **1x** through the `style` option. html-to-image renders the element at its styled size into the bigger canvas, so the diagram ended up in the top-left quarter of the image with the rest empty. The Word report was hit by the same thing because `captureDiagramImage` (used for the docx embed) had the exact same code.

Two more things turned up while digging:

- The exports were **accidentally transparent**. React Flow's `.react-flow` container has `background-color: transparent` by default and the app never sets a background on it — the white you see on screen comes from the page body, which is *outside* the captured element. So every PNG we ever exported had an alpha background, which is a big part of why they looked so empty/broken in viewers.
- If a node was selected when you exported, its **selection border and delete button got baked into the image**, because those are rendered inside the node's DOM.

## The fix

In `frontend/src/features/dfd-editor/lib/export-diagram-image.ts`, both capture blocks now pass the **1x** bounds as `width`/`height` and use `pixelRatio: 2` instead. html-to-image scales the canvas itself, so the output is still 2x resolution but the content fills it. The existing viewport reposition/restore logic, the padding, and both function signatures were deliberately left alone — `exportDiagramImage` just grew one optional `options` param (`backgroundColor`, `excludeIds`), so existing callers didn't need touching. The docx path (`captureDiagramImage`) only gets the crop fix; the Word report code itself is untouched.

## The dialog

One new file: `frontend/src/features/dfd-editor/components/ExportImageDialog.tsx`. All the new behavior lives there, built from the ui primitives already in the repo (dialog, radio-group, checkbox) — no new dependencies.

- **Format**: PNG / SVG radio.
- **Transparent background**: since the old default was accidentally transparent, this got inverted on purpose — unchecked now paints solid white (`backgroundColor: '#ffffff'`), checked passes nothing so you get real alpha. Works for SVG too.
- **Export only selected elements**: builds an `excludeIds` set of everything *not* selected. Selected edges drag their endpoint nodes in so an edge is never drawn dangling, and edges between two exported nodes come along automatically. The checkbox is disabled with a hint when nothing is selected.
- **Selection handling**: before capturing, the dialog deselects everything through the React Flow store (`unselectNodesAndEdges()`), and restores the selection afterwards with `type: 'select'` change events. This matters for two reasons: it keeps selection borders/delete buttons out of the image, and `select`-type changes are exactly the kind both editors' state hooks ignore for undo history and unsaved-changes tracking — using `setNodes` instead would have created garbage undo entries and triggered autosave.
- **Error handling**: Export is disabled on an empty canvas, and a failed export shows a toast and keeps the dialog open instead of closing like it worked.

## Wiring (small diffs)

- `DiagramToolbar.tsx` — the PNG/SVG dropdown became a plain button that opens the dialog. Both editors render this toolbar, so one dialog covers the signed-in editor and the guest editor.
- `DFDEditor.tsx` / `GuestDFDEditor.tsx` — the export handlers accept the options and **return** the export promise. The return matters: the dialog awaits it so the selection restore happens after the capture, not in the middle of it.
- `GuestLayout.tsx` — the `exportImageRef` type widened to match (guest routes the export call through a ref).

## The data-id additions

Four files got a single `data-id={id}` attribute: `DataFlowEdge.tsx`, `TrustBoundaryEdge.tsx`, `CanvasNodeWrapper.tsx`, `GuestNodeWrapper.tsx`.

Reason: edge labels and the red threat-count badges are rendered through React Flow's `EdgeLabelRenderer`, which portals them into a separate DOM layer — outside the edge's own `<g data-id>` element. The selected-only export filters DOM elements by `data-id`, so without these attributes an excluded edge would disappear but leave its label and badge floating in empty space. React Flow's own node and edge elements already carry `data-id`, so this just extends the same convention to the portaled bits.

## What was intentionally not done

- No rewrite of `withExportViewport` to the newer React Flow capture recipe — the existing structure works once the size mismatch is fixed, and a smaller diff is easier to review.
- No draw.io-style zoom %, border width, or shadow options — can be added to the dialog later if anyone asks.
- No changes to reports, the Word helpers, or the backend.

## Testing

`tsc` build and eslint clean on all touched files. Manually verified: tight cropping with white background by default, real transparency when checked (PNG and SVG), selected-only exports without orphaned labels or selection artifacts, selection restored after export, no unsaved-changes flag or undo pollution from exporting, and the guest Word report embedding the corrected image.

vikram-s-narayan (approved):

Issue #246, opened 2026-07-27
Data flow labels are not very readible on the diagram


aganariman:
Dataflow labels are very light grey color and it's difficult to read them on the diagram:

<img width="1134" height="784" alt="Image" src="https://github.com/user-attachments/assets/410b3957-caaa-492b-8e0f-16efad9bb4df" />

Can be more darker text color.

vikram-s-narayan:
Updated version looks like this. 

 
Before:
<img width="322" height="252" alt="Image" src="https://github.com/user-attachments/assets/ccda9218-27fc-49de-8412-6dfed337ee72" />

After:

<img width="346" height="266" alt="Image" src="https://github.com/user-attachments/assets/821d59f5-32ea-4195-8f68-fea426f917b9" />

PR #309, opened 2026-08-12
fix: dataflow labels washed out over system context boundaries

### LINKED ISSUE #307: Dataflow text is lighter when it's displayed over a system context boundary
When a system context element is added to a diagram, color of dataflow labels are lighter compared to when they are used inside a trust boundary.

See image below:
Image

This is visible on a complex and big diagrams - making those labels less readable.
It's even more visible when a dataflow passes from a system context to outside of it:

Image

A similar issue was previously solved for the trust boundaries under ticket [246](https://github.com/precogly/precogly/issues/246).

KhaledZer:
Closes #307

Turns out this wasn't a label styling problem at all (that part was already fixed in #247). React Flow draws edge labels *underneath* nodes, so the system context boundary's background was literally painting over the labels. The boundary used `bg-slate-50/50` — basically a 50% white veil — hence the washed out text. Trust boundaries never had this issue because their backgrounds are only 10% alpha, which also explains why #246's fix looked fine inside them but not here.

So the fix is one line: swap `bg-slate-50/50` for `bg-slate-500/10` on the boundary. Same slate tint, but at the same 10% alpha the trust zones use. And since the component is shared, the guest editor, read-only viewers and image exports all get it for free.

Tested locally in both editors — labels inside a boundary and crossing its edge now read the same as on open canvas. Build passes.

### DIFF
--- a/frontend/src/features/dfd-editor/components/nodes/SystemScopeNode.tsx
+++ b/frontend/src/features/dfd-editor/components/nodes/SystemScopeNode.tsx
@@ -92,7 +92,7 @@ export const SystemScopeNode = memo(function SystemScopeNode({
 
       <div
         className={cn(
-          'w-full h-full rounded-lg border-2 border-solid bg-slate-50/50 transition-all',
+          'w-full h-full rounded-lg border-2 border-solid bg-slate-500/10 transition-all',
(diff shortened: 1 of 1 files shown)


KhaledZer:
image
image

PR #313, opened 2026-08-13
fix: resolve name-collision bugs in CycloneDX guest editor round-trip…

### LINKED ISSUE #311: When more than one component exists with a same name - their threats merge after save and re-open (critical - showstopper)
**Reproduction scenario:**
1. Add two components with a same name (or just copy first one) 
2. Add a new threat (threat 1) to the first component
3. Add another threat (threat 2) to the second component
4. Save the model

Image

5. Start a new threat modeling session
6. Open previously saved threat model
7. Note that first component now has 2 threats

Image

Reproduction model:
[bug reproduction.json](https://github.com/user-attachments/files/30999569/bug.reproduction.json)

If I look into the json file created, I see that threats are associated to components based on their labels, not based on their bom-ref value.

It's important to be able to have multiple elements with a same name in a threat model. It's common to copy paste elements like data stores.
View of the model and threats location should not change when a saved model is opened again.

vikram-s-narayan:
Closes #311                                                                                                                       
                                                                                                                                    
  ## Problem                                                                                                                        
                                                                                                                                    
  The CycloneDX guest editor round-trip (save and re-open) used name-based lookups in several places. When multiple components, data
   flows, threats, or countermeasures shared the same name, identities would collide, causing threats to merge and countermeasures
  to become orphaned.

  ## Changes

  All in `frontend/src/features/guest-editor/lib/cyclonedx-guest.ts`:

  1. **`buildBomRefToNodeIdMap`**: Replaced `type:label` keyed map with positional matching. The Nth blueprint entry maps to the Nth
   matching visualization node/edge, which is guaranteed by serialization order.

  2. **`reconstructCountermeasures`**: Replaced `threatNameToId` (keyed by threat name) with `threatBomRefToGuestId` (keyed by
  abstract threat bom-ref), built from the scenarios array by both callers.

  3. **Serialization `controls.find()`**: Built a `countermeasureIdToControl` map during control construction. Replaced two
  `controls.find(ctrl => ctrl.name === c.name)` calls (abstract threat mitigations and the fix-up loop) with direct ID-based lookup.

  ## Test plan

  - [ ] Create two same-named components, add different threats to each, save, re-open. Threats stay on their respective components.
  - [ ] Create two same-named data flows with threats, save, re-open. Threats stay separate.
  - [ ] Create two same-named threats with different countermeasures, save, re-open. Countermeasures stay linked to the correct
  threat.
  - [ ] Create one threat with two same-named countermeasures, save, re-open. Both remain linked.
  - [ ] Normal round-trip (no duplicate names) still works correctly.

### DIFF
--- a/frontend/src/features/guest-editor/lib/cyclonedx-guest.ts
+++ b/frontend/src/features/guest-editor/lib/cyclonedx-guest.ts
@@ -231,6 +231,8 @@ export function serializeGuestToCycloneDx(
   blueprint.visualizations = [visualization]
 
   // --- Build controls from countermeasures ---
+  // Map countermeasure ID → control object for direct lookup (avoids name collisions)
+  const countermeasureIdToControl = new Map<string, CycloneDxControl>()
   const controls: CycloneDxControl[] = countermeasures.map((c) => {
     const bomRef = refGen.generate('control', c.name || 'control')
     const control: CycloneDxControl = {
@@ -245,6 +247,7 @@ export function serializeGuestToCycloneDx(
     const threatRef = threatIdToBomRef.get(c.threatId)
     if (threatRef) control.mitigations = [threatRef]
 
+    countermeasureIdToControl.set(c.id, control)
     return control
   })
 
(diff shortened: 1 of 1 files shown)


KhaledZer:
I executed the test plan manually and everything worked as expected, the issue seems fixed and all test scenarios passed

vikram-s-narayan:
@KhaledZer - thank you for your help in testing this.

Issue #306, opened 2026-08-12
OAuth tokens keep working after an account is disabled


AlvinKuruvilla:
# OAuth tokens keep working after an account is disabled

Disabling a user does not revoke their OAuth access tokens. The account keeps reading the API
until each token expires. The same account, presenting the JWT the product's own login
issues, is refused immediately.

`is_active` is reachable only through Django's admin, as a superuser; the product has no
control for it. How access gets taken away more generally is on #221. This issue is the
disagreement between the two authentication classes.

Measured against a seeded stack at `c53f4d3`, one account, one request, two credentials:

```
                        account enabled     account disabled
OAuth bearer token      200, 2 rows         200, 2 rows
JWT (simplejwt)         200, 2 rows         401 "User is inactive"
```

`DEFAULT_AUTHENTICATION_CLASSES` holds both classes, so which answer a disabled account gets
depends on which credential it happens to present.

## Why

`is_active` is checked by every authentication class in the stack except the OAuth one.

- `rest_framework_simplejwt.authentication.JWTAuthentication.get_user` raises
  `AuthenticationFailed("User is inactive")`, gated on `CHECK_USER_IS_ACTIVE`, default on.
- `rest_framework.authentication.TokenAuthentication.authenticate_credentials` raises
  `AuthenticationFailed("User inactive or deleted.")`.
- `oauth2_provider.oauth2_validators.OAuth2Validator.validate_bearer_token` does
  `request.user = access_token.user` and checks nothing about the account.
  `AccessToken.is_valid()` is `not self.is_expired() and self.allow_scopes(scopes)`.

django-oauth-toolkit is not wrong to leave it out — `is_valid()` answers "is this token a
live grant", and disabling a user neither expires the grant nor narrows its scope. The check
belongs one layer up, in authentication, which is where the other two libraries put it.

## The same gap reaches the MCP endpoint

The MCP server being built under #221 verifies bearer tokens by reading
`oauth2_provider.AccessToken` directly — it runs in this process, so it needs no call to
`/o/introspect/` — and then acts as the user the token names. It inherits this exactly: a
disabled account's token would keep answering tool calls.

It is patched there, in the verifier that resolves the token, since that is its
authentication layer. Filing this so the REST API gets the same check.

## Suggested fix

Mirror what simplejwt already does, in an `OAuth2Authentication` subclass:

```python
# apps/core/authentication.py
from oauth2_provider.contrib.rest_framework import OAuth2Authentication
[...]

PR #322, opened 2026-08-15
feat(mcp): serve an authorized MCP endpoint, and bring precogly-mcp in-tree

AlvinKuruvilla:
Supersedes the original scope of this PR. It was the authorization-server half;
it now also carries the MCP endpoint and the `precogly-mcp` repository, imported
in-tree. `precogly/precogly-mcp` is to be archived with a pointer to `mcp/`.

## The problem

An MCP client has to act for a user without holding that user's password, and
Precogly had no way to delegate. Separately, the MCP server lived in a second
repository joined to this one by a Docker mount — docker-compose mounted
`../precogly-mcp` and the dev image ran `pip install -e` on it at container
start. That only works for someone who has both repositories checked out as
siblings, which is a setup step nothing states and CI cannot have.

## What this does

**Precogly becomes an OAuth 2.1 authorization server.** django-oauth-toolkit
provides the authorization code flow with PKCE, browser consent, and dynamic
client registration, so a client registers itself, sends the user to a consent
screen, and receives a token scoped to what the user agreed to. Three settings
are non-default and each one blocks a client outright — `DCR_ENABLED`,
`DCR_REGISTRATION_PERMISSION_CLASSES`, and `ALLOW_LOCALHOST_LOOPBACK`; the
commit that sets them explains what breaks without each.

**The MCP endpoint is served from this WSGI process** at `/mcp`, with tokens
verified against django-oauth-toolkit's tables rather than over introspection.
Sharing a process is what removes the credential a separate resource server
would need provisioned into every self-hosted deployment. `mcp/docs/0008` has
the full argument and the two topologies it discards.

**`precogly-mcp` becomes a uv workspace member** at `mcp/`, imported from
precogly/precogly-mcp@6186510. Everything under `mcp/src` and `mcp/tests` is
byte-identical to that commit. `mcp/docs/0011` records why in-tree rather than a
submodule, and what else was rejected.

## What reviewers should look at hardest

- **`backend/config/mcp_mount.py`.** a2wsgi plus a hand-started ASGI lifespan is
  infrastructure in the request path for every MCP call. The lifespan task now
  has an explicit module-level reference: asyncio holds tasks weakly and the
  task's own references form a cycle, so without one the cyclic collector can
  reclaim it and every later request fails with "Task group is not initialized".
- **`OAuth2Authentication` is first in `DEFAULT_AUTHENTICATION_CLASSES`.**
  simplejwt's `get_validated_token` raises rather than returning `None`, so
  anything ahead of it ends the chain for every opaque OAuth token. Reordering
  these breaks MCP with no failing test.
- **The audience check in `apps/core/mcp.py`** is hand-written rather than
  `allows_audience()`, which returns `True` for a token carrying no resource
  indicator at all and matches by URL prefix.

## Migration for contributors

- `uv.lock` moves from `backend/` to the repository root. One workspace, one
  lock.
- The backend image builds from the repository root, not `./backend`. Both
  compose files are updated; `docker compose up` needs nothing else.
- A sibling `../precogly-mcp` checkout is no longer used or needed.

## Verification

Run on the branch at `e90c75b` on 2026-08-29:

```
docker compose run --rm backend python -m pytest -q   ->  273 passed
cd mcp && uv run pytest -q                            ->   46 passed
cd mcp && uv run mypy src                             ->  Success: no issues found in 5 source files
manage.py check --deploy --tag security               ->  no issues (2 silenced)
pre-commit run --from-ref main --to-ref HEAD          ->  passed
uv lock --check                                       ->  Resolved 136 packages
docker compose build backend                                             ->  ok (dev target)
docker compose -f docker-compose.yml -f docker-compose.prod.yml build    ->  ok (prod target)
```

All seven CI checks are green on `e90c75b`.

`precogly-mcp` installs editable inside the image — `direct_url.json` reports
`{"url":"file:///app/mcp","dir_info":{"editable":true}}` — so the compose mount
over `/app/mcp` is live without a rebuild.

## Not done here

- **Revocation controls are deliberately out of scope.** The follow-up I posted
  on #221 (08-12) laid out three options; all three need UI surfaces that do not
  exist, and option 1 runs into the open account-model question in #209. What this
  PR does ship: `/o/revoke_token/` works and cascades both directions, and replay
  detection plus a 7-day inactivity expiry bound a stolen refresh token so it is no
  longer indefinite. What it does not: removing an organization member does not
  touch tokens, `is_active = False` still works at `/api/` (#306), and deleting the
  user is the only reliable revocation. The cheapest next step is navigation rather
  than code — `/o/authorized_tokens/` and its delete view are already routed and
  styled here, and simply unlinked from the SPA.
- **Verified server-side only.** Everything so far is Django's test client,
  which never runs a client's loopback callback listener and never performs a
  real cross-origin redirect. Pointing a real MCP client at a local instance is
  the next step.
- **`/o/register/` is not rate-limited.** Dynamic client registration is open by
  necessity — a client registers before any browser opens, so there is no
  session to authenticate it with. It grants nothing on its own, but it is a
  public write endpoint and wants a limit before an instance faces the internet.
- **The five server-rendered templates are still five.** The second sign-in is
  deliberate (`mcp/docs/0007`); the open question is the Django templates and the
  second CSS build target, which only go away if consent moves into the SPA. The
  2026-08-29 amendment to `mcp/docs/0009` says what actually gates that: not the
  authorization code, which stays in django-oauth-toolkit behind one `OAuthLibMixin`
  call, but what would authenticate a consent POST — a DRF view authenticates with a
  JWT, and a JWT proves someone signed in up to an hour ago, which is the thing 0007
  refuses. The `TODO` in `backend/config/urls.py` records it.


vikram-s-narayan:
Hi @AlvinKuruvilla - this is awesome! I hope to understand and respond on Monday.

AlvinKuruvilla:
Thanks @vikram-s-narayan appreciate it. I'm still trying to wrap my head around parts of it myself 😂 it's surprisingly subtle. Definitely lots of improvements I think we can make and better factoring overall

vikram-s-narayan:
@AlvinKuruvilla - we're having a CI /backend failing check. Per GitHub copilot, the cause is: _"The most likely issue is the mismatch between pytest's configured testpaths: apps and the actual location backend/apps/."_

This may be related to the MCP path.

Do we plan to move the MCP Server to this repo? I see the open issue - https://github.com/precogly/precogly/issues/374 and am wondering if moving that here will help us resolve the failing CI test.

AlvinKuruvilla:
Yeah the plan is to move the MCP server. Im not sure if it will be too much if we include it in this PR. Ill check the CI issue and fix it regardless though

AlvinKuruvilla:
Hi @vikram-s-narayan, I think this is ready for an initial review

vikram-s-narayan (approved):
@AlvinKuruvilla pushed 3 commits, one of which fixes a regression this PR introduces.
Moving the build context to the repo root moved BASE_DIR from /app to /app/backend, so the libraries path derived from it stopped matching the compose mount at /libraries. seed hit the missing path, printed one error line, and carried on to report "Seed complete!", so a fresh docker compose up --build gave you sample threat models with empty diagrams and zero rows in every library catalog. On main the two happened to line up, so this only shows up here.
The three commits: LIBRARIES_PATH becomes one setting instead of a path search, with compose mounting to the default location and seed aborting rather than continuing (plus tests); libraries/ gets copied into the image and removed from .dockerignore, which is a separate older problem since prod runs with no volumes and never had packs at all; and lifespan is annotated as AsyncGenerator rather than AsyncIterator, which Pylance deprecates on @asynccontextmanager.
Heads up: the first commit also reformats 19 pre-existing spots in backend/apps/packs/tests.py. That drift was already on main, but the file was not in this PR's diff before, so the lint job never saw it. Ruff, pytest and mypy all pass locally.

PR #327, opened 2026-08-15
fix: show data assets linked to threat-model flows

### LINKED ISSUE #203: Data Flow: Cannot link Data Assets to a connection
When a **Data Flow** (connection) is created between two components, the properties panel provides a **Data Assets** section with a "Link Asset" option. However, attempting to link a data asset to the data flow does not work. The asset does not get linked/attached to the flow.

**Expected behaviour**:

Clicking "**Link Asset**" under Data Assets should allow the user to select and successfully attach a data asset to the Data Flow connection.

**Actual behaviour**:

The "**Link Asset**" action does not link the data asset to the Data Flow, even though the option is present and clickable.

**Steps to reproduce**:

- Create two components (e.g. "Object Storage" and "Compute Service").
- Draw a **Data Flow** connection between them.
- Open the Data Flow's properties panel.
- Scroll to the Data Assets section and click "**Link Asset**".
- Attempt to select/link a data asset. Observe that it does not get linked to the flow.

Screenshot / GIF:

Image

angelikakasia:
Closes #203

The data asset link was being created, but the API filtered it out when the flow’s source component belonged directly to a threat model instead of an orgsystem.

This updates the queryset to support both ownership paths.

Added API tests to confirm:
- linked assets are returned for threat-model components
- assets from other organizations remain hidden

Testing:
- 2 tests passed
- manually linked multiple assets
- confirmed links and protection settings remained after refresh

### DIFF
--- a/backend/apps/systems/tests/test_data_flow_assets.py
+++ b/backend/apps/systems/tests/test_data_flow_assets.py
@@ -1,0 +1,90 @@
+"""API regression tests for data assets linked to data flows."""
+
+from django.contrib.auth import get_user_model
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from apps.organizations.models import Organization, OrganizationMember
+from apps.systems.models import DataAsset, DataFlow, DataFlowAsset, OrgsystemComponent
+from apps.threat_models.models import ThreatModel
+
+User = get_user_model()
+
+
+class DataFlowAssetAPITests(APITestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.organization = Organization.objects.create(name="Acme")
+        cls.other_organization = Organization.objects.create(name="Globex")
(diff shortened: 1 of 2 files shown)


vikram-s-narayan:
@angelikakasia - thank you. 

  I pushed an additional commit (e82e480) that fixes the same class of bug in 4 other viewsets in views.py:                         
                                                                                                                                    
  - OrgsystemComponentViewSet and ComponentDataAssetViewSet — leaked data across tenants when orgsystem was NULL (the               
  Q(orgsystem__isnull=True) branch had no org check)                                                                                
  - DataFlowViewSet and TrustBoundaryViewSet — excluded records entirely when components belonged to a threat model without an
  orgsystem

  All 4 fixes follow the same pattern from your original fix: check both the orgsystem and threat_model paths to reach the
  organization. Full test suite passes (154 tests).

PR #336, opened 2026-08-19
chore(ci): gate pull requests on Django's deployment checklist

AlvinKuruvilla:
Runs `manage.py check --deploy` as a blocking check.

Two flags do the work:

- `--fail-level WARNING` — every deployment finding is a WARNING and the check framework only fails on ERROR, so the job would otherwise exit 0 whatever it found.
- `--tag security` — a bare run reports 120 issues, 118 of them drf-spectacular schema warnings unrelated to deployment.

The job generates a throwaway `SECRET_KEY` and a placeholder `ALLOWED_HOSTS`, which are environment-supplied in a real deployment. So this guards the settings module, not a running install.

Nothing to fix today: `config/settings/production.py` already sets the whole checklist.

Verified: passes against production settings; fails with 5 findings and exit 1 against `config.settings.development`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,3 +21,24 @@ jobs:
           node-version: '22'
       - run: cd frontend && npm ci
       - run: cd frontend && npx tsc --noEmit
+
+  # Django's deployment checklist, run against the production settings module.
+  deploy-check:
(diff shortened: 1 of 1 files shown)


vikram-s-narayan (approved):
thank you. As always I learn something new with your PRs!

Issue #377, opened 2026-08-23
Move Precogly's Python tooling from pip to uv


AlvinKuruvilla:
**Context:** uv has become the standard Python workflow and the MCP server already uses it. Keeping the backend on pip means two toolchains in one tree once the MCP server folds in. A 0.4.0 dependency. From the 2026-08-22 sync.
**Done looks like:** The backend builds, tests, and Docker images run through uv; requirements files are replaced by a lockfile; contributor docs and CI use it.

From the [2026-08-22 maintainer sync](https://github.com/precogly/precogly/discussions/373).

PR #382, opened 2026-08-23
chore(ci): scan for secrets with gitleaks

AlvinKuruvilla:
A `Security` workflow scanning the PR's own commits, plus a pre-commit hook scanning staged content.

## Why a separate workflow

Triggers are per-workflow, not per-job. A scheduled re-scan — gitleaks' rules improve, so unchanged history yields new findings — would otherwise run pytest and tsc on that schedule too. Trivy, osv-scanner and Scorecard land here as well, and they need `security-events: write`, which `ci.yml` should not grant its test jobs.

## Why not gitleaks-action

It requires a `GITLEAKS_LICENSE` secret for org-owned repos. Free, but it puts a third-party signup and a stored secret in a merge gate for nothing the CLI does not do. The CLI is pinned by version and verified by SHA-256.

## The custom rule

gitleaks 8.30.1 detects a Django `SECRET_KEY` in no form at all — verified, including the placeholder this repo ships. For a Django app that is the highest-value secret in the tree: it signs sessions and password-reset tokens.

The entropy floor is measured, not copied from gitleaks' generic rule (3.7):

| entropy | |
|---|---|
| 3.17 | `your-secret-key-here` — `.env.example` |
| 3.86 | `django-insecure-change-me-in-production` — `base.py` |
| **4.5** | **threshold** |
| 5.31 | `openssl rand -base64 48` — the throwaway in `ci.yml` |
| 5.35 | `startproject`'s generated key |

An earlier draft allowlisted the `django-insecure-` prefix. Dropped: `startproject` generates real keys behind it, so the allowlist would have blinded the rule to the value most likely to be committed by accident.

## Verification

History is clean — 397 commits, 18 MB, 3.3s, all refs.

The rule's first run produced two false positives, both in `search/search_index.json` on `gh-pages`. Minified JSON stores a newline as `\` + `n`, neither of which is whitespace, so the value class ran past the documented placeholder and glued it to the next line, scoring 5.25 as one blob. Fixed by excluding backslash from the value class.

Eight cases after that fix:

- Quiet: the `base.py` default; the `.env.example` placeholder; `env("SECRET_KEY")` with no literal; the minified-JSON regression.
- Fires: a real key as a Python literal; in `.env` form; behind the `django-insecure-` prefix; inside minified JSON.

The pre-commit hook runs and does load `.gitleaks.toml` — confirmed by staging a canary key and watching the commit fail.

`origin/main..HEAD` verified against 8.30.1: clean on the branch as-is, exit 1 with a canary `SECRET_KEY` committed in range, exit 0 with the same commit left outside it.

## Not included

- **No re-scan of existing history.** History was scanned whole once (above) and every commit since arrives through this gate. A newer gitleaks finding something in pre-2026-08-22 history would go unnoticed; a scheduled `--all` job is the fix, deferred until other scanners want the same trigger.
- **This job is not a required status check.** Required checks are `backend` and `frontend`; the `main-branch-protection` ruleset defines none. Until `secrets` is added, it runs, goes red, and does not block. That is already true of `lint` and `deploy-check`.
- No suppression policy — nothing to suppress yet. `.gitleaksignore` if that changes.
- `pip-audit` (waiting on the uv migration, #377), Trivy, osv-scanner, Scorecard, action SHA pinning.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,0 +1,55 @@
+name: Security
+
+on:
+  pull_request:
+    branches: [main]
+
+# Scanners live here rather than in ci.yml so that workflow stays the quality
+# gate — tests, types, lint. The two have different failure meanings: a CI
+# failure says the change is broken, a Security failure says the change is
+# dangerous. Keeping them apart also means a scanner needing wider permissions
+# or a longer checkout does not widen the whole quality gate.
+permissions:
+  contents: read
+
+jobs:
+  # Secrets that reach a public repository are burned the moment they are
+  # pushed, so the only useful place to catch them is before the merge.
+  secrets:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+        # Scan the whole repo history (all commits)
+          fetch-depth: 0
+
+      # Installed from the release archive rather than through
+      # gitleaks/gitleaks-action. The action requires a GITLEAKS_LICENSE secret
+      # for repositories owned by an organization, which precogly/precogly is.
+      # The license is free, but it puts a third-party signup and a stored
+      # secret in the path of a merge gate, and buys nothing the CLI does not
+      # already do. The CLI has no such requirement.
+      #
+      # Bump both values together — the checksum comes from the release's own
+      # gitleaks_<version>_checksums.txt.
+      - name: Install gitleaks
+        env:
+          GITLEAKS_VERSION: 8.30.1
+          GITLEAKS_SHA256: 551f6fc83ea457d62a0d98237cbad105af8d557003051f41f3e7ca7b3f2470eb
+        run: |
+          set -euo pipefail
+          archive="gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz"
+          curl -sSfL -o "$archive" \
+            "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/${archive}"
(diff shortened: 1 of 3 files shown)


.github/workflows/security.yml
```diff
@@ -0,0 +1,55 @@
...
+          sudo install -m 0755 gitleaks /usr/local/bin/gitleaks
+
+      # --log-opts=--all walks every ref, not just the PR head.
+      #
+      # Failing here blocks the merge, which is affordable only because the
+      # history is clean as of 2026-08-22: the first failure will be a real
+      # finding, not a backlog. No suppression policy yet — there is nothing to
+      # suppress. Use `.gitleaksignore` if that changes before one exists.
+      - name: Scan history for secrets
+        run: gitleaks git . --log-opts=--all --redact --verbose --no-banner
```
vikram-s-narayan: I'd like to propose one change:  
replace --log-opts=--all with something scoped to the PR's  commits (e.g. origin/main..HEAD). The full-history scan already served its purpose. You confirmed
all 397 commits are clean. Re-running it on every PR will just get slower as the repo grows, and a gitleaks version bump could fail unrelated PRs by flagging old commits.

AlvinKuruvilla: Thanks for the suggestion, @vikram-s-narayan. I had it that way originally in case we ever change the rules and want to rescan all history, but I acknowledge it's excessive. The most recent commit should address this

AlvinKuruvilla:
@vikram-s-narayan to re-review post addressing feedback

Issue #403, opened 2026-08-27
Stock mitre-atlas and mitre-attack-ics taxonomies contain entries that do not reconcile against MITRE


navesecurity:
## Summary

Two of Precogly's stock taxonomies carry entries that do not reconcile against the authoritative upstream source: `mitre-atlas` contains three technique identifiers absent from MITRE ATLAS, and `mitre-attack-ics` contains six techniques MITRE has revoked plus one with the wrong title. Because these are stock content, any pack that maps to them inherits the error, and a threat model citing them will not survive review against MITRE's own data.

This is the same class as #339, where a stock compliance pack did not correspond to a real publication.

## 1. Three `mitre-atlas` entries are not in MITRE ATLAS

Checked against `mitre-atlas/atlas-data` release **5.6.0** (`dist/ATLAS.yaml`, 170 techniques), which is the machine-readable source MITRE publishes. Its highest identifier is `AML.T0112` with sub-techniques `.000` and `.001`. Precogly carries 173 entries, three of which have no counterpart:

| Precogly entry | Title |
|---|---|
| `AML.T0091.001` | Use Alternate Authentication Material: Web Session Cookie |
| `AML.T0113` | Steal Web Session Cookie |
| `AML.T0114` | AI Service Web Interface |

Every one of the 170 official techniques **is** present in Precogly, so this is three additions rather than a version skew in the other direction. If they come from a newer unreleased ATLAS revision, recording that provenance on the entries would resolve it; if not, they should be withdrawn.

## 2. Six `mitre-attack-ics` techniques are revoked upstream

Checked against `mitre/cti` `ics-attack/ics-attack.json`. MITRE restructured these in the recent ICS revision and recorded an explicit `revoked-by` relationship for each:

| Precogly entry | Revoked in favour of |
|---|---|
| `T0803` Block Command Message | `T1691.001` Command Message |
| `T0804` Block Reporting Message | `T1691.002` Reporting Message |
| `T0839` Module Firmware | `T1693.002` Module Firmware |
| `T0855` Unauthorized Command Message | `T1692.001` Command Message |
| `T0856` Spoof Reporting Message | `T1692.002` Reporting Message |
| `T0857` System Firmware | `T1693.001` System Firmware |

The successors are not one-for-one on tactics. `T1693.001` spans Persistence, Inhibit Response Function and Impair Process Control, so a migration has to re-read tactics rather than carry them across.

## 3. `T0815` carries the wrong title

[...]

PR #410, opened 2026-08-27
Reconcile taxonomy joins on every sync, not only forced ones

### LINKED ISSUE #409: Non-forced sync of the #403 fix destroys all mitre-attack-ics threat links and rebuilds none
## Summary

Applying the #403 fix to an existing deployment with a **non-forced** `sync_all_packs_from_source()` destroys every `mitre-attack-ics` threat-taxonomy link and rebuilds none of them. On our instance that would have been 60 links across the `ot-ics` library.

The cause is an interaction between two things that are individually reasonable:

1. `#403`'s fix bumps `taxonomies/mitre-attack-ics` 1.0.0 -> 2.0.0, but leaves `threat-libraries/ot-ics` at 1.0.0 while changing `ot-ics/joins/threats-mitre-attack-ics.yaml`.
2. The third pass that re-applies taxonomy joins from disk is gated on `if force:`.

## Walking the non-forced path

`sync_all_packs_from_source()` skips a pack only when `is_in_database and not force and database_version == version`.

- `mitre-attack-ics` **is not skipped**, since 1.0.0 != 2.0.0. It re-imports, and `_import_pack` calls `_hard_delete_pack_items(existing)`, which does `ExternalTaxonomy.objects.filter(source_pack=pack).delete()`. `ThreatLibraryTaxonomyEntry.taxonomy_entry` is `on_delete=CASCADE`, so **every** link into that taxonomy goes, not only the 6 revoked ids.
- `ot-ics` **is skipped**, since its version did not change, so its corrected `threats-mitre-attack-ics.yaml` is never read.
- The third pass at `services.py:1848-1863` would have re-read it, but it sits under `if force:` and does not run.

Net effect: the taxonomy is correct, and nothing points at it.

With `force=True` the third pass runs and the result is correct, which is presumably how the fix was validated.

## Suggested fixes

Either one closes it; the second is the more general.

1. **Bump `threat-libraries/ot-ics`.** It has a content change in this commit and its version did not move, so no resync will read it. This is worth a general rule: a pack whose files change needs its version changed, or deployments silently keep the old content. We added a CI check for exactly this in our own pack repo today after being bitten by it.

2. **Ungate the third pass.** Re-applying taxonomy joins from disk is idempotent, since `_load_threat_taxonomy_joins` writes through `get_or_create`/`update_or_create`, so running it unconditionally costs a re-read and protects against any ordering where a taxonomy is imported after the packs referencing it. We carry this locally as an extracted `reconcile_taxonomy_joins_from_source()` called from both `sync_all_packs_from_source()` and `import_single`, because `import_single` has the same gap: importing one taxonomy pack on its own leaves every referencing pack's joins unre-applied. Happy to open a PR if useful.

## How this was found

Verified against MITRE's own data while reconciling our packs: `mitre-atlas` 5.6.0 `dist/ATLAS.yaml`, and the enterprise/ICS/mobile STIX bundles from `mitre/cti`. The #403 content fix itself checks out. The six removed ICS ids are all genuinely revoked, and their replacements match the `revoked-by` relationships in the ICS bundle:

| removed | revoked-by | name |
|---|---|---|
| T0803 | T1691.001 | Command Message |
| T0804 | T1691.002 | Reporting Message |
| T0839 | T1693.002 | Module Firmware |
| T0855 | T1692.001 | Command Message |
| T0856 | T1692.002 | Reporting Message |
| T0857 | T1693.001 | System Firmware |

We were not hit by the link loss, because we already run the third pass ungated. We did hit the missing `ot-ics` bump and had to force that pack's import by hand.

navesecurity:
Fixes #409.

A taxonomy pack's re-import CASCADE-deletes every threat-taxonomy join pointing into it, so those joins have to be re-read from disk afterwards. That pass exists, but was reachable only two ways, neither of which covers an ordinary upgrade:

- `sync_all_packs_from_source()` ran it under `if force:`
- `import_single` never ran it at all

## The concrete case

#403 bumps `taxonomies/mitre-attack-ics` 1.0.0 → 2.0.0 and changes `threat-libraries/ot-ics/joins/threats-mitre-attack-ics.yaml` without bumping `ot-ics`. On a non-forced sync:

1. `mitre-attack-ics` is **not** skipped, since its version changed. It re-imports, and `_hard_delete_pack_items` does `ExternalTaxonomy.objects.filter(source_pack=pack).delete()`. `ThreatLibraryTaxonomyEntry.taxonomy_entry` is `on_delete=CASCADE`, so **every** link into that taxonomy goes, not only the six revoked ids.
2. `ot-ics` **is** skipped, since its version did not change, so its corrected joins are never read.
3. The pass that would rebuild them sits under `if force:` and does not run.

The taxonomy ends up correct with nothing pointing at it. On our deployment that was 60 links.

## Changes

- Extract the third pass as `reconcile_taxonomy_joins_from_source()` and call it unconditionally at the end of `sync_all_packs_from_source()`. Re-reading is idempotent, since `_load_threat_taxonomy_joins` writes through `get_or_create`/`update_or_create`, so the only cost is the re-read.
- Call it from `_import_pack` after the transaction commits, closing the `import_single` gap. Reading committed data is the point: the pack that resolves a pending join is usually a different pack imported separately, so re-trying inside the same transaction would re-read the snapshot that produced the gap. A failure there is logged rather than raised, since the pack's own import has already succeeded.
- Bump `ot-ics` 1.0.0 → 1.1.0 so its corrected joins are actually read.

## Verification

```
ruff check apps/packs/services.py apps/packs/tests.py   All checks passed!
python -m pytest apps/packs/tests.py -q                 34 passed
```

Both new tests fail against `main` and pass here, so they are a real guard rather than a restatement:

```
FAILED TaxonomyJoinReconciliationTests::test_forced_sync_reconciles_joins
FAILED TaxonomyJoinReconciliationTests::test_non_forced_sync_still_reconciles_joins
2 failed, 32 passed
```

## One note

The requirement-overlay **second** pass is still gated on `if force:` and has the same shape. Left alone to keep this focused, but it looks worth the same treatment.

We have run the extracted function in production since 2026-08-20 and used it again today to reconcile 1978 joins across 38 packs while applying the #403 fix.


vikram-s-narayan:
Bug is real, fix is correct. One change needed:

## Only reconcile when a taxonomy pack is reimported

Right now `reconcile_taxonomy_joins_from_source()` runs after every single-pack import. But the CASCADE only happens when a **taxonomy** pack is force-reimported — threat library and compliance packs don't own `ExternalTaxonomy` rows, so they can never trigger this bug. No reason to scan all 39 packs' join files on every unrelated import.

Guard it:

```python
pack_type = pack_meta.get("pack_type", "")
if pack_type == "taxonomy" and existing and force:
    try:
        reconcile_taxonomy_joins_from_source()
    except Exception:
        logger.exception(
            f"Taxonomy join reconciliation failed after importing pack "
            f"'{slug}'. The pack itself imported successfully; only the "
            f"catalog-wide retry failed."
        )
```

Thank you!

navesecurity:
Thanks — guard applied in 2146520, with one part left off that I want to flag rather than apply silently.

**Scoped to `pack_type == "taxonomy"`.** Your reasoning holds and I checked it against a live catalogue rather than taking it on trust: all ten packs owning `ExternalTaxonomy` rows declare `pack_type: taxonomy`, and no pack of any other type owns one. An ordinary threat or compliance import now skips the pass entirely, which is every import except the four taxonomy packs.

**I have not added `existing and force`,** because I think it would leave half the bug in place.

That pair covers the re-import case, where the CASCADE destroys joins. But `reconcile_taxonomy_joins_from_source()`'s own docstring names two ways a join fails to resolve, and the first one is:

> the taxonomy pack owning the referenced code **has not been imported yet**, or is re-imported afterwards to add the code in question

The first of those is a *first* import: `existing` is `None` and `force` is `False`. Nothing is destroyed, but the referencing pack was imported earlier against a catalogue that did not yet contain the code, so its joins were parked as `PendingTaxonomyOverlay` rows with a warning while the import still reported success. This pass is what settles them. Gating on `existing and force` would leave them parked indefinitely, and the only signal is a warning on an import that said it succeeded.

Concretely: import `medtech-cwe` into an instance that already has a pack citing CWE codes, and the citing pack's joins stay pending until something re-reads them.

Your performance objection is fully met by `pack_type` on its own, so the two conditions are separable. Happy to add them if you still want the tighter scope, but I did not want to narrow past the bug without saying so.

Two tests pin the guard in both directions.

AlvinKuruvilla:
FYI @navesecurity @vikram-s-narayan, I enabled branch protection, so all CI workflows must run and pass before a PR can be merged and reviewed. Update the branch to pull in the new MCP workflow so CI can run it.

vikram-s-narayan:
@navesecurity -thank you for this contribution. @AlvinKuruvilla - thanks for setting up the branch protection.

PR #414, opened 2026-08-28
Embed rendered DFDs in Word reports

### LINKED ISSUE #206: Embed DFD diagram image in signed-in editor Word report
The guest editor now embeds the DFD diagram as a PNG in its Word report (#188). The signed-in editor's Word export still uses a
text placeholder ("Insert DFD diagram screenshot here").                                                                          
                                                            
This wasn't addressed in the PR (https://github.com/precogly/precogly/pull/205) that addresssed issue #188 because the signed-in editor generates its report from the Reports tab, where the React Flow canvas isn't mounted — so there's no DOM element to capture from.

Possible approach to handling this: when the user navigates to the Reports tab, briefly render a hidden/offscreen ReactFlow instance with the
diagram data, capture the image, then tear it down. Alternatively, capture and cache the diagram image on the DFD editor tab
whenever nodes change, and read from that cache when generating the report.

sidd190:
## Summary

- Capture each saved DFD through the existing React Flow renderer.
- Embed the rendered PNG in the Word report instead of generating a simplified outline.
- Preserve the existing fallback for diagrams without nodes.

[Screencast From 2026-08-29 01-34-45.webm](https://github.com/user-attachments/assets/d58d0244-fd5b-475f-b471-093c5684ae08)

Closes #206

### DIFF
--- a/frontend/src/components/shared/ReadOnlyDFDViewer.tsx
+++ b/frontend/src/components/shared/ReadOnlyDFDViewer.tsx
@@ -3,17 +3,18 @@
  * Displays a Data Flow Diagram without editing capabilities.
  */
 
-import { ReactFlow, Background, Controls, ReactFlowProvider } from '@xyflow/react'
+import { forwardRef, useImperativeHandle, useRef } from 'react'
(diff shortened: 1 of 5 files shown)


vikram-s-narayan (approved):
Changes looking good!

PR #416, opened 2026-08-29
Rename selected components by typing or pasting

### LINKED ISSUE #299: Select component and start typing it's name to rename it or paste from clipboard
It would speed up diagramming if it was possible to select a component on the panel and just start typing text that will be applied to it as name.

Also would be great if I could just click Ctrl+V to paste text from clipboard just after selecting a component. Text from clipboard should also directly apply as the component name in this case.

sidd190:
## Summary

- Start renaming a selected component by typing its replacement name.
- Paste clipboard text directly into a selected component with `Ctrl/Cmd + V`.
- Preserve normal node copy/paste behavior for multi-selection and edge selection.
- Use the trusted paste event to avoid browser clipboard permission prompts.

[Screencast From 2026-08-29 12-00-31.webm](https://github.com/user-attachments/assets/52ec2f9c-e320-42c3-9e92-704ddee318f4)

Closes #299

### DIFF
--- a/frontend/src/features/dfd-editor/DFDEditor.tsx
+++ b/frontend/src/features/dfd-editor/DFDEditor.tsx
@@ -373,6 +373,22 @@ function DFDEditorContent() {
     }
   }, [boundaryMode, cancelBoundaryMode])
 
+  const startNodeEditing = useCallback((initialText: string) => {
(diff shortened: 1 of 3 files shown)


vikram-s-narayan (approved):
Really cool feature!

PR #419, opened 2026-08-29
Add resizable sticky note annotations

### LINKED ISSUE #276: Sticky note element
Would be great to have a sticky note element that can be placed anywhere on the diagram.

Ideally it should be possible to select a style on the properties pane when this element is selected - for example picking a sticky paper color.
Also would be great to have some basic text formatting options.

See examples below:

Image

Image

sidd190:
## Summary

Closes #276.

Adds Sticky Note annotations to the DFD editor with a compact, focused editing experience.

## Changes

- Add Sticky Note creation from the diagram toolbar.
- Support Sticky Notes in authenticated and guest editors.
- Keep Sticky Notes independent from DFD data-flow connections.
- Remove connection handles from Sticky Notes.
- Add selected-only resize controls with minimum dimensions of 120×80.
- Persist resized dimensions in the node style.
- Support preset note colors, bold/italic text, and small/medium/large text sizes.

## Validation

- Focused ESLint checks pass for the changed sticky-note files.
- `git diff --check` passes.
- Full type-check is currently blocked by pre-existing locally missing `docx` and `html-to-image` modules; both are already declared in the frontend manifest and lockfile.

Screenshots and manual testing will be added separately.


vikram-s-narayan:
@sidd190 - This is exciting! Can you please add the video of how the feature works as this is a major front-end change?

vikram-s-narayan:
There was a minor merge conflict with the imports in lucide and I have fixed it.

But we still need a little more work before we can merge this:

1. Not sure if we need a description field. Think of a sticky note. It's a blunt instrument for a quick annotation that's all. So I would remove the description AND the name. And simply allow user to enter the text into the "Message" field. The bold and italic features for text is useful and should be retained.

Screenshot 2026-08-31 at 11 30 01 AM

2. Additionally, `stickyNote` is wrapped in `withThreatBadge` in both `CanvasNodeWrapper.tsx` and `GuestNodeWrapper.tsx`, but sticky notes are annotations with no semantic meaning in the threat model. They'll never have threats linked to them, so `useThreatCount` will always return 0. Register them directly as `stickyNote: StickyNoteNode` without the wrapper.

PR #421, opened 2026-08-29
Reconnect dataflow endpoints

### LINKED ISSUE #305: Possibility to change source or destination component of a dataflow
When connecting two components with a dataflow, it can happen that connection is not as desired.
It would be great if it was possible to select a dataflow line, then select either it's start or it's end and drag and re-attach it somewhere else.

As a workaround, obviously we can delete the data flow and then create a new one correctly. But in this case properties of the dataflow are lost and work must be redone.

A simple example:
After reorganizing positions of the elements, there is a wish to re-attach the dataflow from "Plot..." process to the left side of the "Exported drawing file" data store.
Image

This case might be simple, but often it can happen that we want to add a new component and reorganize existing dataflows - moving destination side from old component to a new one.

On draw.io it's when a dataflow line is selected, it has draggable bullets at start and end of the line:
Image

Thanks.

sidd190:
## Summary

- Allow existing dataflows to reconnect to a different source or destination component.
- Preserve the dataflow's existing properties during reconnection.
- Enable the behavior in both authenticated and guest editors.

[Screencast From 2026-08-29 18-45-07.webm](https://github.com/user-attachments/assets/cc4d6880-8606-4d01-ab89-950ff9674262)

## Testing

- Reconnected both source and target endpoints.
- Verified dataflow properties remained intact.
- Tested authenticated and guest editor flows.

Closes #305

### DIFF
--- a/frontend/src/features/dfd-editor/DFDEditor.tsx
+++ b/frontend/src/features/dfd-editor/DFDEditor.tsx
@@ -11,6 +11,7 @@ import {
   type Connection,
   type XYPosition,
   addEdge,
+  reconnectEdge,
 } from '@xyflow/react'
 import '@xyflow/react/dist/style.css'
 import { ArrowLeft, Save, Clock, Loader2, Pencil, Trash2, ShieldAlert } from 'lucide-react'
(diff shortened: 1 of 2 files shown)


vikram-s-narayan (approved):
sweet!

Issue #405, opened 2026-08-27
[Security bug] : GET /api/frameworks/ returns other tenants' internal standards


AlvinKuruvilla:
# `GET /api/frameworks/` returns other tenants' internal standards

`StandardFramework` holds two kinds of row in one table, and the field that decides which
says so itself (`backend/apps/compliance/models.py:22`):

```python
threat_model = models.ForeignKey(
    "threat_models.ThreatModel",
    ...
    help_text="Populated for user-created internal standards. "
    "NULL for global pack-sourced frameworks.",
)
```

Rows with `threat_model` null are shipped reference data — NIST, ISO, SOC 2 — and every
tenant is meant to read them. Rows with it set belong to one threat model, and so to one
organization.

`StandardFrameworkViewSet` (`backend/apps/compliance/views.py:20`) serves both from a
class-level `queryset = StandardFramework.objects.all()` with no `get_queryset`.
`IsSecurityTeam` returns `True` for safe methods, so nothing narrows the read.

## Reproduction

```python
home_org = Organization.objects.create(name="Kestrel", domain="kestrel.test")
other_org = Organization.objects.create(name="Contoso", domain="contoso.test")

outsider = User.objects.create_user(
    username="outsider", email="outsider@kestrel.test", password="testpass123"
)
OrganizationMember.objects.create(organization=home_org, user=outsider, role="member")

# Shared reference data: `threat_model` null, every tenant reads it.
StandardFramework.objects.create(
    slug="soc2", name="SOC 2", version="2017", issuer="AICPA"
)

# One tenant's own standard, attached to that tenant's threat model.
other_model = ThreatModel.objects.create(name="Contoso ledger", organization=other_org)
StandardFramework.objects.create(
    slug="contoso-internal",
    name="Contoso Internal Control Standard",
    version="1.0",
    issuer="Contoso",
    threat_model=other_model,
)

client = APIClient()
client.credentials(
    HTTP_AUTHORIZATION=f"Bearer {RefreshToken.for_user(outsider).access_token}"
)
```

As a member of Kestrel only:

```
GET /api/frameworks/  ->  200
                          ["Contoso Internal Control Standard", "SOC 2"]
```

Expected: `SOC 2` present, `Contoso Internal Control Standard` absent. Written as two
assertions, the reference-data half passes and the tenant half fails.

## Fix

Narrowing the queryset closes the read, and has to keep serving the null-`threat_model`
rows to everyone.

[...]

Issue #258, opened 2026-07-30
Imports select the organization implicitly, so a multi-org user's import can land in the wrong tenant


AlvinKuruvilla:
`import_tm_library` and the CycloneDX import both resolve the target organization with:

```python
first_membership = request.user.organization_memberships.first()
```

`apps/threat_models/views.py:896` and `:975`.

`organization_memberships` has no explicit ordering, so `.first()` returns whichever row Postgres yields. For a user in one organization that is always right. For a user in several it is arbitrary, and the import — a bulk write creating a threat model with its components, flows, threats and countermeasures — lands in whichever tenant the database happened to return.

## Effect

- A multi-org user can create a threat model in an organization they did not choose.
- Nothing in the response says which organization received it.
- Silent: no error is raised, and the model simply appears missing from the organization the user was working in.

## Fix

The adapters already accept `organization` as a parameter (`apps/threat_models/adapters/base.py:10`), so this is threading an explicit value through the view rather than resolving it from the membership set.

Suggested shape: accept an optional `organization` in the request, validate it against the caller's memberships, and return 400 when the caller has more than one membership and did not specify one.

## Notes

Agreed as a blocker in discussion #221 ("MCP for Precogly"), where an agent's token may cover several organizations and the write lands wherever `.first()` points.

Deliberately not labelled `Security`: the caller is a legitimate member of every organization involved, so this is data integrity rather than privilege escalation. #226 and #227 are genuine cross-tenant leaks and this should not dilute that signal.

Issue #227, opened 2026-07-19
[Security bug] : Null-orgsystem analysis components leak across tenants and allow cross-tenant component-data-asset writes


sidd190:
## Summary

Analysis-only components can have `orgsystem = null` while still being linked to a threat model. The components queryset includes null-orgsystem components without requiring the linked threat model to belong to the current user's organization.

This lets a user from Org A read analysis-only components belonging to Org B. A related endpoint is worse: `ComponentDataAsset` associations for those null-orgsystem components can be updated by the wrong-org user.

## Impact

This is a cross-tenant data exposure and write vulnerability.

An authenticated user can see another organization's analysis-only components and associated data assets. They can also modify component-data-asset metadata such as `encrypted` and `volume`, corrupting another tenant's model data.

The issue affects threat-model integrity because analysis-only components are part of the threat model scope even when they are not attached to an orgsystem.

## Evidence

Similar to #226, multi org creation through UI is not exposed anyhow, so used curl requests as explained below here

Test setup:

- User A: `sec02-a@example.com`, member of Org A only
- User B: `sec02-b@example.com`, member of Org B only
- Victim threat model: ID `7`, owned by Org B
- Victim analysis-only component: ID `708`, `orgsystem = null`, `threatModel = 7`
- Victim component-data-asset association: ID `1`

Read leak request made as User A:

```http
GET /api/components/?threat_model=7
Authorization: Bearer <Org A user token>
```

Actual result:

```http
200 OK
```

Response excerpt:

```json
{
  "id": 708,
  "name": "SEC-03 Victim Analysis-Only Component",
  "orgsystem": null,
  "threatModel": 7
}
```

Expected result:

The component should not be returned because the caller does not belong to the threat model's organization.

Cross-tenant write request made as User A:

```http
PATCH /api/component-data-assets/1/
Authorization: Bearer <Org A user token>
Content-Type: application/json

{
  "encrypted": true,
  "volume": "edited by wrong org"
}
```

Actual result:

```http
200 OK
```

Response excerpt:

```json
{
  "id": 1,
  "component": 708,
  "componentName": "SEC-03 Victim Analysis-Only Component",
  "dataAsset": 1,
  "dataAssetName": "SEC-03 Victim Data Asset",
  "volume": "edited by wrong org",
  "encrypted": true
}
```

Expected result:

The request should return `403 Forbidden` or `404 Not Found`.

## Root Cause

[...]

Issue #226, opened 2026-07-19
[Security bug] : Cross-tenant DFD creation allows attaching diagrams to another org's threat model


sidd190:
## Summary

The DFD creation endpoint accepts a caller-supplied `threat_model_id` and looks up the threat model globally. It does not scope that lookup to the caller's organizations and does not run object permission checks on the target threat model.

As a result, an authenticated user from Org A can create a DFD attached to a threat model owned by Org B.

## Impact

This is a cross-tenant write vulnerability. A user can attach diagrams to another organization's threat model. DFDs are not harmless metadata: they are part of the core modeling surface, and primary DFDs can drive component/threat-analysis state through the diagram sync flow.

This can corrupt another tenant's model, create misleading diagrams under another org's threat model, and potentially affect downstream report/threat-analysis behavior once the diagram is updated or used.

## Evidence

There's no option to create any org other than the one with admin@precogly.dev seeded in the UI so done solely with curl requests.

Test setup:

- User A: `sec02-a@example.com`, member of Org A only
- User B: `sec02-b@example.com`, member of Org B only
- Victim threat model: ID `7`, owned by Org B

Request made as User A:

```http
POST /api/diagrams/
Authorization: Bearer <Org A user token>
Content-Type: application/json

{
  "name": "SEC-02 Cross Tenant DFD Created By User A",
  "diagram_type": "level1",
  "threat_model_id": 7,
  "canvas_data": {
    "nodes": [],
    "edges": [],
    "viewport": { "x": 0, "y": 0, "zoom": 1 }
  }
}
```

Actual result:

```http
201 Created
```

Response excerpt:

```json
{
  "id": 3,
  "name": "SEC-02 Cross Tenant DFD Created By User A",
  "diagramType": "level1",
  "threatModel": 7,
  "isPrimary": true,
  "updatedBy": 2,
  "updatedByEmail": "sec02-a@example.com"
}
```

Expected result:

The request should return `403 Forbidden` or `404 Not Found` because the caller does not belong to the victim threat model's organization.

## Root Cause

`DFDViewSet.create()` performs a global threat model lookup:

```py
ThreatModel.objects.get(id=threat_model_id)
```

The lookup is not filtered by the request user's organizations. The endpoint also does not call object-level permission checks against the target threat model before creating the DFD.

Affected area:

- `backend/apps/diagrams/views.py`
- `DFDViewSet.create()`

## Suggested Fix

[...]

PR #429, opened 2026-08-30
fix(systems): close the trust zone leaks, and make every model declare its owner

### LINKED ISSUE #404: [Security bug] : ?threat_model= on /trust-zones/ returns other tenants' zones
# `?threat_model=` on `/trust-zones/` returns other tenants' zones

`TrustZoneViewSet.get_queryset` (`backend/apps/systems/views.py:78`) derives the caller's
organizations and then discards them when `threat_model` is present:

```python
def get_queryset(self):
    org_ids = self.request.user.organization_memberships.values_list(
        "organization_id", flat=True
    )
    threat_model_id = self.request.query_params.get("threat_model")
    if threat_model_id:
        # Scoped: zones that have components belonging to this threat model
        return TrustZone.objects.filter(
            components__threat_model_id=threat_model_id
        ).distinct()          # org_ids is unused on this branch
    ...
```

A member of one organization reads another organization's trust zones by passing that
organization's threat model id. Without the parameter the same zone is correctly hidden,
so the default queryset is not at fault — the parameter replaces it rather than narrowing
it.

Same class as #227, different route.

## Reproduction

```python
# Two organizations with no member in common.
home_org = Organization.objects.create(name="Kestrel", domain="kestrel.test")
other_org = Organization.objects.create(name="Contoso", domain="contoso.test")

outsider = User.objects.create_user(
    username="outsider", email="outsider@kestrel.test", password="testpass123"
)
OrganizationMember.objects.create(organization=home_org, user=outsider, role="member")

# A trust zone reachable only through the other organization's threat model.
other_model = ThreatModel.objects.create(name="Contoso ledger", organization=other_org)
other_zone = TrustZone.objects.create(name="Contoso DMZ")
OrgsystemComponent.objects.create(
    name="Contoso edge proxy", trust_zone=other_zone, threat_model=other_model
)

client = APIClient()
client.credentials(
    HTTP_AUTHORIZATION=f"Bearer {RefreshToken.for_user(outsider).access_token}"
)
```

```
GET /api/trust-zones/                                 -> 200, "Contoso DMZ" absent    (correct)
GET /api/trust-zones/?threat_model=<other_model.pk>   -> 200, "Contoso DMZ" present   (leak)
```

## Why the sweep that was aimed at this missed it

e82e480 (`fix: add threat-model org check to 4 viewset querysets`) covered
`OrgsystemComponent`, `ComponentDataAsset`, `DataFlow`, and `TrustBoundary`. It did not
reach `TrustZoneViewSet`, and a review asking "is the base queryset organization-scoped?"
would have passed it — the base queryset *is* scoped. The leak is in an early return that
throws that queryset away, which is only visible if you also ask what each query parameter
does.

Every hand-read query parameter has that property. #266 covers the other eight.

## Fix

Declare `threat_model` on a `FilterSet` so it narrows the organization-scoped queryset
Instead of replacing it, which is what #266 proposes, we can either follow in the fix there or capture this as a one-off fix
## The part a filter cannot fix

`TrustZone` has no foreign key to an organization, and no forward path to one at all —
it is reached only in reverse, through `components__orgsystem__organization` or
`components__threat_model__organization`, either of which may be null. Structurally it sits
with the shared catalog (`ThreatLibrary`, `LibraryPack`, the taxonomies) rather than with
tenant-owned data, and `TrustBoundary` is in the same position.

So nothing prevents components in two different organizations from pointing at the same
zone. When that happens, the row belongs to both and no queryset filter resolves it. Closing
this issue stops the parameter from reaching past the caller's organizations; it does not
give the row an owner. That belongs in a separate design discussion, which is being written
up now.

### LINKED ISSUE #406: [Security bug] : a DFD canvas can rename and reclassify another tenant's trust zone
# A DFD canvas can rename and reclassify another tenant's trust zone

`_sync_nodes_to_trust_zones` (`backend/apps/diagrams/services.py:964`) resolves the
`trust_zone_id` carried in a canvas node and writes the node's label and trust level onto
whatever row comes back:

```python
existing_trust_zone_id = node_data.get("trust_zone_id")

if existing_trust_zone_id:
    try:
        trust_zone = TrustZone.objects.get(id=existing_trust_zone_id)
        trust_zone.name = label
        trust_zone.trust_level = trust_level
        trust_zone.description = description
        trust_zone.save()
```

Canvas data is caller-supplied and the lookup is unfiltered. The sync runs from
`DFDViewSet.perform_update` on the caller's own primary DFD, so no cross-tenant object
access is needed at the view layer — the caller updates a diagram they own, and the write
lands on a trust zone belonging to another organization.

Unlike #404, which is a read on the same model, this one cannot be fixed by scoping a
queryset. `TrustZone` carries no organization, so there is nothing to filter the lookup
against.

## Reproduction

```python
attacker_org = Organization.objects.create(name="Kestrel", domain="kestrel.test")
victim_org = Organization.objects.create(name="Contoso", domain="contoso.test")

attacker = User.objects.create_user(
    username="attacker", email="attacker@kestrel.test", password="testpass123"
)
OrganizationMember.objects.create(
    organization=attacker_org, user=attacker, role="security_team"
)

# The victim's zone, reachable only through the victim's threat model.
victim_model = ThreatModel.objects.create(name="Contoso ledger", organization=victim_org)
victim_zone = TrustZone.objects.create(name="Contoso DMZ", trust_level=20)
OrgsystemComponent.objects.create(
    name="Contoso edge proxy", trust_zone=victim_zone, threat_model=victim_model
)

# The attacker's own threat model and primary DFD. Nothing here is shared.
attacker_model = ThreatModel.objects.create(
    name="Kestrel checkout", organization=attacker_org
)
attacker_dfd = DFD.objects.create(
    name="Kestrel L1",
    threat_model=attacker_model,
    is_primary=True,
    canvas_data={"nodes": [], "edges": []},
)
```

Then, as the attacker:

```http
PATCH /api/diagrams/<attacker_dfd.pk>/
Authorization: Bearer <Kestrel user token>
Content-Type: application/json

{
  "canvasData": {
    "nodes": [
      {
        "id": "zone-1",
        "type": "trustZone",
        "data": {
          "label": "Adopted by Kestrel",
          "trustZoneId": <victim_zone.pk>,
          "trustLevel": 99
        }
      }
    ],
    "edges": []
  }
}
```

```
200 OK

victim_zone.refresh_from_db()
(victim_zone.name, victim_zone.trust_level)
    actual    ('Adopted by Kestrel', 99)
    expected  ('Contoso DMZ', 20)
```

## Impact

`trust_level` is the damaging half. It is the zone's security posture and it feeds threat
analysis, so the victim's model starts producing different answers with no edit to anything
they can see. The rename alone would be defacement; the reclassification is an integrity
bug.

## Fix

Short term, the sync has to constrain the lookup to zones reachable from the DFD being
updated — the ids already present in that diagram's canvas — rather than accepting any id
the caller supplies.

That is a patch, not a fix. The lookup is unfiltered because the model has no owner to
filter on, which is step 2 of `docs/design/tenancy-boundary.md`.

Related: #404 (read on the same model), #226 (cross-tenant DFD creation), #227.

AlvinKuruvilla:
Closes #404. Closes #406.

## The problem

Six cross-tenant defects are open — #226, #227, #404, #405, #406, #258, each
found by hand. e82e480 swept four viewsets for exactly this bug class and still
missed #404.

They keep happening because the boundary is re-derived in every viewset, as a join
from the caller's memberships to whatever the model hangs off. Nothing in the
schema lets anyone check the join. Of the 61 models in `apps/`, 43 reach
`Organization`, 38 of those by more than one path and every one through a nullable
foreign key. `TrustZone`, `TrustBoundary` and `VerificationTest` hold customer data
and reach it by no forward path at all.

## What this does

**Every model in `apps/` declares who its rows belong to** — `TENANT_OWNED`,
`SHARED_REFERENCE` or `MIXED`. 

This PR also adds a Django system check that fails on any model that
doesn't declare its tenancy.
43 are tenant-owned, 12 are shared reference, and 6 are mixed.

**`TrustZone` and `TrustBoundary` are a bit of a special case and get a non-null `organization`.**
Trust zones had no column saying which organization they belong to. Anything that needed to know walked out through the zone's components to a threat model or a system, and from there to an organization — several hops, more than one route, and a nullable link on most of them. Each viewset picked its own route, and nothing checked the choice.

Two bugs came out of that:

- `GET /api/trust-zones/?threat_model=` returned other organizations' zones (#404). The viewset joined through two of the available paths and they were the wrong two.
- `PATCH /api/diagrams/<your own dfd>/` overwrote another organization's zone (#406). The canvas sync reads a `trust_zone_id` out of the request body, looks it up with no filter, and writes the node's name and trust level onto whatever comes back. Trust level feeds threat analysis, so the row is left wrong, not just renamed. There was no queryset to fix here — the lookup was unfiltered because there was nothing to filter on.

This adds a non-null `organization` to `TrustZone` and `TrustBoundary`. It duplicates nothing on these two: `TrustZone` had no foreign key reaching an organization at all.

The migration derives each existing zone's owner once, through its components. 
**There are two edge cases we still need to handle in the future:**
1. A zone reachable from two organizations (which is #406 having already happened) 
2. A zone reachable from none. 
For right now, both raise and name the rows, and nothing is deleted. `manage.py trust_zone_owners` prints the same list so we can look before you migrate.

## Verification

Run on the branch at `dd09b5e` on 2026-08-30:

```
uv run pytest -q                                       ->  242 passed
manage.py check                                        ->  no issues (0 silenced)
manage.py makemigrations --check --dry-run             ->  No changes detected
pre-commit run --from-ref upstream/main --to-ref HEAD  ->  passed
```

Run with uv against a local Postgres; this branch predates the compose rework in
#322 and the cached image no longer matches its mounts. All six CI checks are green
on `dd09b5e`, which includes `docker compose run --rm backend python -m pytest`.

Both regression tests were checked against the unfixed code. The #404 pair fails
without the viewset change; #406 reproduces as
`('Adopted by Kestrel', 99) != ('Contoso DMZ', 20)`.

## Not done here

- **Six `# noqa: DJ001` on the branch; two of them are deferrals.** The four on
  `qualified_slug` are permanent — `null=True` is required by the unique constraint,
  since Postgres treats NULLs as distinct and `""` would collide. The two that are
  debt are `Risk.response` and `OrgsystemComponent.category`. Both need a data
  migration, and neither is only that: `category` feeds a `.tm` export branch no test
  covers (`tm_library.py:1362` exports a component as an actor when the category is
  NULL), and `response` is what the risk board filters on with `null` as a column
  key, so backend and frontend have to land together. No issue filed for either.

### DIFF
--- a/backend/apps/ai/models.py
+++ b/backend/apps/ai/models.py
@@ -14,6 +14,7 @@ from django.conf import settings
 from django.db import models
 
 from apps.core.models import TimestampedModel
+from apps.core.tenancy import Tenancy
 
 from .crypto import decrypt, encrypt
 from .providers.base import ResolvedConfig
@@ -21,6 +22,8 @@ from .providers.base import ResolvedConfig
 
 class AIProviderConfig(TimestampedModel):
     """An organization's configuration for one AI model endpoint."""
+
+    tenancy = Tenancy.TENANT_OWNED
 
     class ProviderType(models.TextChoices):
         # Only OpenAI-compatible endpoints are supported today. New, non-
@@ -114,6 +117,8 @@ class AIUsageRecord(TimestampedModel):
     system. ``model``/``provider_type`` are snapshots so history stays correct
     when a config is later edited or deleted.
     """
+
+    tenancy = Tenancy.TENANT_OWNED
 
     organization = models.ForeignKey(
         "organizations.Organization",
--- a/backend/apps/compliance/models.py
+++ b/backend/apps/compliance/models.py
@@ -5,11 +5,16 @@ Compliance models - frameworks, standards.
 from django.db import models
 
 from apps.core.models import TimestampedModel
+from apps.core.tenancy import Tenancy
 from apps.threats.models import CountermeasureLibrary
 
 
 class StandardFramework(TimestampedModel):
     """Compliance framework (e.g., PCI-DSS, SOC2, NIST)."""
+
+    # The user-created internal standards are unscoped today: every tenant reads
+    # them — precogly/precogly#405.
+    tenancy = Tenancy.MIXED
 
     slug = models.SlugField(max_length=100, unique=True)
     source_pack = models.ForeignKey(
@@ -43,6 +48,10 @@ class StandardFramework(TimestampedModel):
 class StandardRequirement(TimestampedModel):
     """Requirement within a compliance framework."""
 
+    # No nullable key of its own; a requirement belongs to whoever its framework
+    # belongs to.
+    tenancy = Tenancy.MIXED
+
     framework = models.ForeignKey(
         StandardFramework,
         on_delete=models.CASCADE,
@@ -73,6 +82,10 @@ class StandardRequirement(TimestampedModel):
 
 class CountermeasureLibraryStandard(models.Model):
     """Association between countermeasure and compliance requirement."""
+
+    # Both ends are library rows, so the mapping ships with the packs rather than
+    # belonging to whoever happens to use it.
+    tenancy = Tenancy.SHARED_REFERENCE
 
     class Sufficiency(models.TextChoices):
         FULL = "full", "Full"
@@ -109,6 +122,9 @@ class StandardRequirementMapping(models.Model):
     Example: NIST CSF PR.AC-4 partially covers OWASP A01:2021.
     Used for gap analysis between compliance standards.
     """
+
+    # A statement about two published standards, true for every installation.
+    tenancy = Tenancy.SHARED_REFERENCE
 
     class Sufficiency(models.TextChoices):
         FULL = "full", "Full"
--- a/backend/apps/compliance/models.py
+++ b/backend/apps/compliance/models.py
@@ -98,7 +98,9 @@ class CountermeasureLibraryStandard(models.Model):
         unique_together = ["countermeasure_library", "requirement"]
 
     def __str__(self):
-        return f"{self.countermeasure_library} -> {self.requirement} ({self.sufficiency})"
+        return (
+            f"{self.countermeasure_library} -> {self.requirement} ({self.sufficiency})"
+        )
 
 
 class StandardRequirementMapping(models.Model):
--- a/backend/apps/core/apps.py
+++ b/backend/apps/core/apps.py
@@ -5,3 +5,7 @@ class CoreConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "apps.core"
     verbose_name = "Core"
+
+    def ready(self):
+        # Importing registers the checks; nothing else uses the module.
+        from apps.core import checks  # noqa: F401
--- a/backend/apps/core/checks.py
+++ b/backend/apps/core/checks.py
@@ -1,0 +1,80 @@
+"""Startup checks that hold conventions the type system cannot.
+
+Registered from `CoreConfig.ready`.
+"""
+
+from django.apps import apps as django_apps
+from django.core.checks import Error, register
+
+from apps.core.tenancy import Tenancy
+
+# Only Precogly's own models are asked to declare. Django's own tables, allauth's, and
+# every other dependency's are outside the question — nothing here scopes them and
+# nothing here may change them.
+DECLARING_APPS = frozenset(
+    {
+        "ai",
+        "compliance",
+        "core",
+        "diagrams",
+        "organizations",
+        "packs",
+        "systems",
+        "threat_models",
+        "threats",
+    }
+)
+
+
+@register()
+def every_model_declares_its_tenancy(app_configs, **kwargs):
+    """Refuse to start if a model in `apps/` has not said who its rows belong to.
+
+    The point is the models that do not exist yet. A new model added without a
+    `tenancy` attribute fails here on the first `manage.py` command rather than
+    reaching review as an unscoped queryset nobody thought to ask about — which is how
+    #404 and #405 arrived.
+
+    Inheritance is checked deliberately with `vars()` rather than `getattr`: a concrete
+    model inheriting another concrete model would otherwise pick up its parent's answer
+    silently, and the parent's answer is not evidence about the child.
+
+    `app_configs` is honoured rather than ignored, so `manage.py check organizations`
+    reports on that app alone. It is also the seam the tests use: under `isolate_apps`
+    the models being checked are not in the global registry, and a check that only ever
+    read `django_apps` could not see them.
+    """
+    if app_configs is None:
+        models = django_apps.get_models()
+    else:
+        models = [model for config in app_configs for model in config.get_models()]
+
+    errors = []
+
+    for model in models:
+        if model._meta.app_label not in DECLARING_APPS:
+            continue
+
+        declared = vars(model).get("tenancy")
+        if isinstance(declared, Tenancy):
+            continue
+
+        if declared is None:
+            hint = (
+                "Add `tenancy = Tenancy.TENANT_OWNED` if every row belongs to one "
+                "organization, or `Tenancy.SHARED_REFERENCE` if every installation "
+                "holds the same rows. See apps/core/tenancy.py."
+            )
+        else:
+            hint = f"`tenancy` is {declared!r}; it must be a `Tenancy` member."
+
+        errors.append(
+            Error(
+                f"{model._meta.label} does not declare its tenancy.",
+                hint=hint,
+                obj=model,
+                id="precogly.E001",
+            )
+        )
+
+    return errors
--- a/backend/apps/core/tenancy.py
+++ b/backend/apps/core/tenancy.py
@@ -1,0 +1,102 @@
+"""Whether a model's rows belong to one organization or to every installation.
+
+Precogly serves several organizations from one database, and the rule keeping them
+apart has been written once per viewset — as a join from the caller's memberships to
+whatever the model happens to hang off. Measured across the 61 models in `apps/`, 38
+of the 43 that reach `Organization` do so by more than one path, all 38 through a
+nullable foreign key somewhere along the way, and three models that hold customer
+data — `TrustZone`, `TrustBoundary`, `VerificationTest` — reach it by no forward path
+at all. So each of those joins is a guess, nothing checks it, and
+six cross-tenant defects have been found by hand (#226, #227, #404, #405, #406, #258).
+
+The fact that no join can recover is which kind of row a table holds. That is what
+this module records, and `apps.core.checks` is what makes recording it compulsory.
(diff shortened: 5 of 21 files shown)


vikram-s-narayan:
@AlvinKuruvilla - the FK fix on TrustZone/TrustBoundary look solid. The two leaks are clearly real and the fix is the right approach.

I have a question about the tenancy declaration system (TENANT_OWNED/SHARED_REFERENCE/MIXED + the system check). I'm not sure what bug class it prevents that the schema doesn't already handle:

- Every tenant-owned model already reaches Organization through its FK chain. The two that didn't (TrustZone, TrustBoundary) are fixed in this PR by adding the FK directly.
- Non-null FKs already prevent creating records without an org at the database level.
- The annotation catches "forgot to add a label," but not "forgot to filter by org in the viewset," which is the actual bug pattern behind all six issues.

Could you help me understand the scenario you're guarding against?

AlvinKuruvilla:
Hi @vikram-s-narayan 
**TL;DR** — You're right that this PR doesn't catch "forgot to filter by org." It's 61 annotations plus a check that guards them: it says which models are tenant-owned, step 2 gives exactly those models a non-null `organization` column, step 3 filters every query on that column.

"Forgot to filter in the viewset" describes two of the six defects, #227 and #404. The other four aren't that bug: #226 leaks from `create()` on a viewset whose `get_queryset` is correct; #406 has no viewset to fix; #258 picks the wrong organization with every `NOT NULL` satisfied; #405 can't be filtered at all, because ownership there is a per-row fact.

I created a tracking issue so we don't forget the context and can have a place to recod decisions or changes: #441 
_____
This PR does not catch "forgot to filter by org." What landed here is 61 annotations plus a check that guards them. It says which models are tenant-owned; step 2, which would be a separate PR, gives exactly those models a non-null `organization` column; step 3, another PR most likely, filters every query on that column, refusing to serve one it can't scope.

Four of the six defects are a different bug.

## #227 and #404 are the ones you're describing

Both have a `get_queryset` that scopes incorrectly, and both are fixed by writing a better join.

## #226: the viewset filters correctly and leaks anyway

`apps/diagrams/views.py:42-50` scopes the list:

```python
return DFD.objects.filter(threat_model__organization_id__in=org_ids)
```

Twelve lines below, `create()` on the same viewset:

```python
threat_model_id = request.data.get("threat_model_id")        # :65
threat_model = ThreatModel.objects.get(id=threat_model_id)   # :73
```

Nobody forgot. `get_queryset` is right there; DRF doesn't call it on this path. Still live on this branch.

## #406: there is no viewset to fix

`apps/diagrams/services.py:964` on `main`:

```python
trust_zone = TrustZone.objects.get(id=existing_trust_zone_id)
trust_zone.name = label
trust_zone.trust_level = trust_level
trust_zone.save()
```

Reached from `perform_update` on the caller's own DFD, with `trust_zone_id` read out of caller-supplied canvas JSON. No `get_queryset` exists in this file. Worse, on `main` there was nothing to filter against — `TrustZone` carried no organization. The fix on this branch is `apps/diagrams/services.py:1033`:

```python
trust_zone = TrustZone.objects.get(id=existing_trust_zone_id, organization_id=org_id)
```

That predicate became writable only because 772edf0 added the column. The bug was neither a missing filter nor a missing label. It was a missing fact.

## #258: the filter is irrelevant, the organization is just wrong

`apps/threat_models/views.py:990`:

```python
first_membership = request.user.organization_memberships.first()   # :990
organization = first_membership.organization                        # :996
threat_model, summary = adapter.import_data(json_data, organization, request.user)  # :1020
```

`.first()` on an unordered queryset. A multi-org user's import lands in an arbitrary one of their organizations. Every `NOT NULL` is satisfied. The row is in the wrong tenant.

A constraint says a row has an owner. It says nothing about which owner, and nothing about who may read it. None of the six defects is "a row was created without an organization."

The numbers don't support the premise either. Walking forward FK paths from every tenant-owned model on this branch: of 42, 32 have a `NOT NULL` path to `Organization`, 9 reach it only across a nullable hop, and one has no forward path at all. `OrgsystemComponent` is the clearest: `orgsystem` (`apps/systems/models.py:352`), `trust_zone` (`:369`) and `threat_model` (`:383`) are all `null=True`, so the database permits a component belonging to nobody.

## #405: the filter cannot be written

`apps/compliance/models.py:27-34`:

```python
threat_model = models.ForeignKey(
    "threat_models.ThreatModel", on_delete=models.CASCADE,
    null=True, blank=True, related_name="internal_frameworks",
    help_text="Populated for user-created internal standards. "
              "NULL for global pack-sourced frameworks.",
)
```

`apps/compliance/views.py:23` is a `ModelViewSet` with no `get_queryset` override at all:

```python
queryset = StandardFramework.objects.all()
```

Adding one doesn't resolve it, because some rows in that table are supposed to be world-readable. The filter has to know which, and the only thing that says so is `threat_model IS NULL` — a key kept for attachment, read as ownership. Five more of the same shape: `apps/threats/models.py:25`, `:101`, `:220`, `apps/systems/models.py:201`, `apps/diagrams/models.py:27`, every one `null=True` under `help_text="Pack this item came from (null = custom or legacy)"`.

No foreign key chain reaches a per-row fact.

## On reachability as the predicate

Your first point is true of nearly everything else too, which is why it can't be the test. BFS from each model to `Organization`, run on this branch:

```text
--- BFS: forward + reverse ---
  shared-reference models the predicate calls tenant-owned:  12/12

--- BFS: forward only ---
  shared-reference models the predicate calls tenant-owned:   2/12
      compliance.CountermeasureLibraryStandard
      compliance.StandardRequirementMapping
  tenant-owned models the predicate calls exempt:
      threats.VerificationTest
```

Including reverse relations, reachability calls the entire shared catalogue tenant-owned. Restricting to forward keys, it exempts `VerificationTest`, which holds one organization's `passed`, `evidence` and `last_run_at`.

`VerificationTest`'s viewset does everything right and still cannot hold the line. `apps/threats/views.py:894-903`:

```python
test_ids = InstanceCountermeasureTest.objects.filter(
    countermeasure__threat_model__organization_id__in=org_ids
).values_list("verification_test_id", flat=True)
return VerificationTest.objects.filter(id__in=test_ids)
```

`InstanceCountermeasureTest` (`apps/threats/models.py:702-720`) constrains `unique_together = ["countermeasure", "verification_test"]` and nothing else, so two organizations' countermeasures may point at one `VerificationTest`. When they do, `id__in=test_ids` returns it to both and the `ModelViewSet` lets both write `passed` and `evidence`. Same shape as #406, in a viewset that filters. Flagged at `apps/threats/models.py:677-681`, unverified — declaring every model is what turned it up.

Does that clear up the intention? To track these steps, I created tracking issue #441

PR #457, opened 2026-09-03
make dependabot less noisy with PRs

vikram-s-narayan:
Dependabot makes too many PRs - many of which are noisy. This reduces their frequency to a more manageable level.


AlvinKuruvilla (approved):
Looks good to me

PR #464, opened 2026-09-03
Sync frontend CountermeasureStatus with the backend's 8 values (fixes #328)

### LINKED ISSUE #328: CountermeasureStatus frontend type/config missing 3 backend enum values (in_progress, implemented, decommissioned)
## What happened

Clicking into a threat whose linked countermeasures included one set to `implemented`, `in_progress`, or `decommissioned` crashed the DFD editor's threat panel:

```
TypeError: Cannot read properties of undefined (reading 'color')
    at renderItem (ComponentView.tsx:1061)
    at SortableItem (SortableList.tsx:9:99)
```

## Root cause

The Django backend's `InstanceCountermeasure.status` field has 8 valid choices: `gap`, `planned`, `in_progress`, `implemented`, `verified`, `waived`, `platform`, `decommissioned`.

The frontend's `CountermeasureStatus` type (`frontend/src/features/dfd-editor/types/threat-analysis.ts`) only ever covered 5 of those:

```ts
export type CountermeasureStatus = 'platform' | 'gap' | 'planned' | 'verified' | 'waived'
```

`COUNTERMEASURE_STATUS_CONFIG` is keyed by that narrower type, and `ComponentView.tsx:1061` does an unguarded lookup:

```ts
const statusConfig = COUNTERMEASURE_STATUS_CONFIG[cm.status]
// ...
style={{ backgroundColor: statusConfig.color }}
```

Any countermeasure with a status the frontend doesn't know about (`implemented`, `in_progress`, `decommissioned`, all valid backend values) produces `statusConfig === undefined`, and `.color` throws.

`deriveThreatStatus()` (same file) has a similar, non-crashing gap: it explicitly checks for `gap`/`planned`/`waived` and falls through to `'mitigated'` for anything else, so `implemented`/`decommissioned` land in "mitigated" by default (arguably fine) but `in_progress` should logically land in "addressable" and doesn't.

## Suggested fix

Widen the frontend type and config to match the backend enum exactly, and add `in_progress` to the "addressable" bucket in `deriveThreatStatus`. We patched this locally (not yet a PR):

```ts
export type CountermeasureStatus =
  | 'platform' | 'gap' | 'planned' | 'in_progress'
  | 'implemented' | 'verified' | 'waived' | 'decommissioned'

export const COUNTERMEASURE_STATUS_CONFIG: Record<CountermeasureStatus, {...}> = {
  // ...existing 5...
  in_progress: { label: 'In Progress', color: '#eab308', bgColor: 'bg-yellow-500', description: 'Implementation actively underway' },
  implemented: { label: 'Implemented', color: '#10b981', bgColor: 'bg-emerald-500', description: 'Implemented, not yet independently verified by the security team' },
  decommissioned: { label: 'Decommissioned', color: '#6b7280', bgColor: 'bg-gray-400', description: 'Control formally retired, no longer tracked as active' },
}
```

## Secondary, non-crashing gap in the same area

`CountermeasureStatusButtons.tsx` only renders 5 clickable status buttons (the same 5 the type used to support), so there's currently no way to *set* a countermeasure to `implemented`/`in_progress`/`decommissioned` from the UI, only via direct API access. Doesn't crash, just incomplete once the type is widened.

## Reproduction

1. Set any `InstanceCountermeasure.status` to `implemented` (or `in_progress`/`decommissioned`) via the API or admin.
2. Open the DFD editor, click a threat linked to that countermeasure.
3. `ComponentView.tsx` throws on render.

Found while doing a full evidence-grounded countermeasure implementation-status assessment on a real threat model (99 records, several legitimately needed `implemented` rather than only `gap`/`verified`).

navesecurity:
`InstanceCountermeasure.Status` on the backend has carried 8 values — `gap`, `planned`, `in_progress`, `implemented`, `verified`, `waived`, `platform`, `decommissioned` — since `0012_cyclonedx_schema_enrichment` (2026-07-11). The frontend never caught up: `CountermeasureStatus` in `dfd-editor/types/threat-analysis.ts` still had only 5, so `COUNTERMEASURE_STATUS_CONFIG[status]` returned `undefined` for any countermeasure the backend had marked `in_progress`, `implemented` or `decommissioned`. That is the crash reported in #328.

This widens `CountermeasureStatus` to the full 8-value union, adds display config (label/color/description) for the 3 missing statuses, and updates `deriveThreatStatus()` so `in_progress` counts toward *addressable* alongside `planned`/`waived`, and `implemented`/`decommissioned` join `platform`/`verified` under *mitigated*.

It also fixes a second, independent copy of the same drift: `threat-models/api/threats.ts` declared its own local `BackendCountermeasureStatus` as a 5-value literal union, commented "aligned with frontend CountermeasureStatus" but kept in sync only by hand. That is now a direct type alias to the imported type, so the two can no longer diverge.

Verified with a full `tsc -b --noEmit` in the frontend container: 0 errors after both changes. The first change alone fails the build (TS2322 in `useWorkspaceThreatAnalysis.ts`, `in_progress` not assignable to the stale union), which is what shows the second copy was load-bearing.

Fixes #328

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/src/features/dfd-editor/types/threat-analysis.ts
+++ b/frontend/src/features/dfd-editor/types/threat-analysis.ts
@@ -18,7 +18,7 @@ export interface ComplianceStandardMapping {
 /**
  * Status of a countermeasure for a specific component-threat
  */
-export type CountermeasureStatus = 'platform' | 'gap' | 'planned' | 'verified' | 'waived'
+export type CountermeasureStatus = 'platform' | 'gap' | 'planned' | 'in_progress' | 'implemented' | 'verified' | 'waived' | 'decommissioned'
 
 export const COUNTERMEASURE_STATUS_CONFIG: Record<
   CountermeasureStatus,
@@ -53,6 +53,24 @@ export const COUNTERMEASURE_STATUS_CONFIG: Record<
     color: '#3b82f6', // blue
     bgColor: 'bg-blue-500',
     description: 'Risk accepted, not implementing',
+  },
+  in_progress: {
+    label: 'In Progress',
+    color: '#eab308', // yellow
+    bgColor: 'bg-yellow-500',
+    description: 'Implementation actively underway',
+  },
+  implemented: {
+    label: 'Implemented',
+    color: '#10b981', // teal-green, distinct from verified
+    bgColor: 'bg-emerald-500',
+    description: 'Implemented, not yet independently verified by the security team',
+  },
+  decommissioned: {
+    label: 'Decommissioned',
+    color: '#6b7280', // gray
+    bgColor: 'bg-gray-400',
+    description: 'Control formally retired, no longer tracked as active',
   },
 }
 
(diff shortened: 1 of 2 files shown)


vikram-s-narayan (approved):
This is awesome. Thank you for catching and fixing!

Issue #318, opened 2026-08-14
Force-reimporting a pack silently orphans live threat models (component library links, threat library links, and cross-check-flagged duplicates)


navesecurity:
# Force-reimporting a pack silently orphans live threat models (component library links, threat library links, and cross-check-flagged duplicates)

## Summary

Force-reimporting any library pack that a threat model is already using (via the "Sync" action in Libraries, or `import_pack_from_path(..., force=True)`) silently corrupts every threat model built from that pack. Three distinct symptoms, all from the same root cause:

1. `OrgsystemComponent.component_library` goes `NULL` on every component sourced from the reimported pack.
2. `ComponentInstanceThreat.threat_library` / `DataFlowInstanceThreat.threat_library` go `NULL` on every threat sourced from the reimported pack.
3. Manually regenerating threats afterward (the natural recovery step) creates a second, correctly-linked set of threats *alongside* the orphaned originals rather than replacing them, so the threat model ends up with duplicates as well as orphans.

None of this produces an error, a warning, or a log message. It was only caught because a downstream feature (the pack cross-check panel, "N threats not backed by pack relationships") happened to flag the duplicates by count.

## Root cause

`_hard_delete_pack_items()` in `apps/packs/services.py` (~line 1635) hard-deletes and recreates `ComponentLibrary`, `ThreatLibrary`, and `CountermeasureLibrary` rows on every force-reimport, with new surrogate IDs:

```python
def _hard_delete_pack_items(pack: LibraryPack):
    """Hard delete all library items from a pack.

    Note: Instance models use SET_NULL for library FKs, so deleting library items
    will orphan instances but not delete them. This preserves user work.
    """
    ComponentLibrary.objects.filter(source_pack=pack).delete()
    ThreatLibrary.objects.filter(source_pack=pack).delete()
    CountermeasureLibrary.objects.filter(source_pack=pack).delete()
    ...
```

The docstring's claim ("this preserves user work") is only half true:

- `OrgsystemComponent.component_library`, `ComponentInstanceThreat.threat_library`, and `DataFlowInstanceThreat.threat_library` are all `on_delete=SET_NULL`, so the instance *rows* survive, but their link back to the (now newly-recreated) library entry does not get re-established anywhere in the reimport path.
[...]

PR #465, opened 2026-09-03
Preserve compliance mappings when a requirement is reimported: CASCADE to SET_NULL (fixes #338)

### LINKED ISSUE #338: CASCADE delete on compliance requirement reimport silently drops cross-pack countermeasure mappings
## Summary

Reimporting a compliance-framework pack whose `pack.yaml` renames or typos a `section_code` silently CASCADE-deletes every `CountermeasureLibraryStandard` mapping (and any `StandardRequirementMapping` cross-framework mapping) that pointed at the old `StandardRequirement` row — including mappings created by *other* packs that happen to reference that requirement. This is the third instance of the same "grouped delete/reimport operation silently drops cross-references living outside the immediately-touched pack" pattern as #318 and #329.

## Root cause

Two separate problems compound:

**1. Inconsistent `on_delete` policy.** `backend/apps/compliance/models.py`:
- `CountermeasureLibraryStandard` (class at line 74): both FKs use `on_delete=models.CASCADE` — `countermeasure_library` (line 83) and `requirement` (line 88).
- `StandardRequirementMapping` (class at line 104): both `from_requirement` (line 117) and `to_requirement` (line 122) also use `on_delete=models.CASCADE`. Only its `source_pack` FK (line 132) uses `SET_NULL`.

This is the opposite of the deliberate pattern used everywhere else an instance references a library object — e.g. `OrgsystemComponent.component_library`, `ComponentInstanceThreat.threat_library`, `DataFlowInstanceThreat.threat_library` all use `SET_NULL`, specifically so that "deleting library items will orphan instances but not delete them" (docstring of `_hard_delete_pack_items()`, `backend/apps/packs/services.py` lines 1643-1648 — the same function #318's local patch disables).

**2. Automatic purge of "stale" requirements on every reimport.** `_load_frameworks()`, `backend/apps/packs/services.py`, lines 2205-2216:

```python
# Purge stale requirements no longer in the pack YAML
incoming_section_codes = {
    req.get("section_code", "")
    for req in framework_data.get("requirements", [])
}
incoming_section_codes.discard("")

stale_count, _ = StandardRequirement.objects.filter(
    framework=framework
).exclude(
    section_code__in=incoming_section_codes
).delete()
```

This runs on *every* `import_single` reimport of a compliance pack, not just force-reimports. Any `StandardRequirement` whose `section_code` doesn't appear in the incoming pack YAML gets deleted — whether it was genuinely retired or just renamed/typo'd.

**Combined effect:** renaming or typo-ing one `section_code` in a compliance pack's YAML, then reimporting via `import_single`, deletes the `StandardRequirement` row, which CASCADEs to delete every `CountermeasureLibraryStandard` row referencing it (from *any* pack that mapped a countermeasure to that requirement, not just the reimported one) and every `StandardRequirementMapping` row referencing it (cross-framework gap-analysis links, again potentially sourced from other packs). The only trace left behind is a `logger.info("Purged N stale requirements...")` line, which says nothing about the cascade to countermeasure/cross-framework mappings.

## Relationship to #318 / #329

Third instance of the same "grouped delete/reimport operation silently drops cross-references that live outside the immediately-touched pack" pattern:
- **#318**: force-reimport `SET_NULL`s `component_library`/`threat_library` FKs, orphaning instances with no re-link path.
- **#329**: TM-Library export groups multi-instance threats and exports only one representative instance's data, silently dropping the rest.
- **This issue**: compliance-pack reimport CASCADE-deletes a `StandardRequirement`, which cascades further to delete cross-pack `CountermeasureLibraryStandard`/`StandardRequirementMapping` rows.

This one is more severe than #318 in one respect: #318 orphans instance data (recoverable — the row survives in the DB with a `NULL` FK); this genuinely deletes rows with no `SET_NULL` fallback. It's only recoverable via `sync_from_source(force=True)`'s extra requirement-overlay re-application pass, not via a plain `import_single` reimport of the affected pack — a plain reimport of the pack that originally created the mapping won't recreate it unless that pack's own current YAML still declares it, and if the mapping came from a *different* pack than the one whose `section_code` changed, that other pack's own reimport won't touch the deleted requirement at all.

## Suggested repro

1. Pick (or create) a compliance pack requirement that has a `CountermeasureLibraryStandard` mapping from another pack's countermeasure.
2. Count `CountermeasureLibraryStandard.objects.filter(requirement__section_code="<code>").count()` — note it's > 0.
3. Rename that requirement's `section_code` in the pack's YAML (typo or legitimate rename), leaving everything else unchanged.
4. Reimport via `import_single`.
5. Re-run the count from step 2 — now 0. The mapping is gone, not orphaned, not flagged, and not recreated under the new `section_code` unless the source pack that owns the mapping also redeclares it on its own next reimport.

## Suggested fix

Same fix family as #318: change `CountermeasureLibraryStandard.requirement`, `StandardRequirementMapping.from_requirement`, and `StandardRequirementMapping.to_requirement` to `on_delete=models.SET_NULL` (with `null=True`), matching the pattern used for every other instance-facing FK in the codebase. Separately, `_load_frameworks()` should warn explicitly (not just an info-level log) when a purge is about to cascade-delete non-empty `CountermeasureLibraryStandard`/`StandardRequirementMapping` sets, so a renamed/typo'd `section_code` doesn't fail silently.

## Environment

- Precogly commit `f5be72e1` (`backend/apps/compliance/models.py`, `backend/apps/packs/services.py`), confirmed via direct read of the live source, 2026-08-20.

navesecurity:
Reimporting a compliance pack deletes and recreates its `StandardRequirement` rows. Three FKs pointed at those rows with `on_delete=CASCADE` — `CountermeasureLibraryStandard.requirement`, `StandardRequirementMapping.from_requirement` and `.to_requirement` — so every reimport silently destroyed the mappings pointing into them, including the full/partial sufficiency judgement a security engineer had recorded, with no trace they had existed.

This switches all three to `SET_NULL, null=True`, matching the design already used one model over for the same relationship: `InstanceCountermeasureStandard.requirement` in `apps/threats/models.py` has always been `SET_NULL`.

Every consumer of these fields was audited and null-guarded rather than left to treat orphaned rows as real mappings:

- `apps/packs/services.py` — `get_active_overlays_for_pack()`
- `apps/compliance/serializers.py` — `CountermeasureLibraryStandardSerializer`
- `apps/threat_models/compliance_service.py` — drift/sync functions
- `apps/diagrams/services.py` — instance-mapping propagation

The trade-off is the same one #318 makes: an orphaned mapping lingers as a harmless, meaningless row rather than being silently destroyed. This model keeps no `section_code`/name snapshot, so such a row cannot describe what it used to point at. That is clutter, but strictly better than losing the judgement call with no record.

308 lines of tests cover the reimport path and each null-guarded consumer.

One note, and I am happy to be corrected if I have misread it: #338 is currently marked closed, but `apps/compliance/models.py` on `main` still has all three fields on `CASCADE`. It looks like the issue was closed without the code change landing.

Fixes #338

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vikram-s-narayan (approved):
Thank you for taking care of this. I also appreciate the clear comments. It helps us understand better.

PR #466, opened 2026-09-04
fix: scope DFD creation to caller's organization

vikram-s-narayan:
DFDViewSet.create() resolved the threat_model_id from the request body with an unscoped ThreatModel.objects.get(id=...), which meant any authenticated user could attach a DFD to any organization's threat model by supplying its ID. This also triggered the auto-primary logic (is_first_dfd), so against a threat model with no existing diagrams, the injected DFD became the primary DFD driving the assessment. The 404/success split additionally served as an existence oracle for threat model IDs across organizations.

Three other actions in the same viewset (ai_availability, analyze_image, generate_dfd) already scope the threat model lookup to the caller's organizations. This fix applies the same pattern to create().

Changes

- Replaced the unscoped ThreatModel.objects.get(id=...) in DFDViewSet.create() with ThreatModel.objects.filter(id=..., organization_id__in=org_ids).first()
- Added regression tests covering own-org success, cross-org rejection (404), nonexistent ID, auto-primary for own org, and cross-org primary injection prevention

Credit

Thanks to @navesecurity for finding this vulnerability and responsibly disclosing it.

### DIFF
--- a/backend/apps/diagrams/tests/test_dfd_create_org_scoping.py
+++ b/backend/apps/diagrams/tests/test_dfd_create_org_scoping.py
@@ -1,0 +1,73 @@
+"""Regression test: DFDViewSet.create() must scope threat model to the caller's orgs.
+
+Without org scoping, any authenticated user can attach a DFD to any threat model
+by supplying its id, regardless of organization membership.
+"""
+
+from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


AlvinKuruvilla (approved):
Looks good to me, LGTM!

PR #510, opened 2026-09-09
add vercel tracking for frontend to track app.precogly.org

vikram-s-narayan:
Set up tracking on Vercel so that we can see visitors for app.precogly.org
Per my understanding, Vercel Web Analytics is cookieless and collects no personally identifiable information. It counts page views and visitors using anonymized, hashed identifiers that aren't stored or linked back to individuals. No cookie consent banner is required for it under GDPR.

We've set this up such that if an org is not deploying on Vercel, it does not affect them (requires you to setup VITE_ENABLE_ANALYTICS  = true on Vercel).

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -26,6 +26,7 @@
         "@radix-ui/react-tabs": "^1.1.13",
         "@radix-ui/react-tooltip": "^1.2.8",
         "@tanstack/react-query": "^5.90.12",
+        "@vercel/analytics": "^2.0.1",
         "@xyflow/react": "^12.10.0",
         "class-variance-authority": "^0.7.1",
         "clsx": "^2.1.1",
@@ -1292,9 +1293,6 @@
         "arm"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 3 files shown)


AlvinKuruvilla:
Nice, this looks good to me. It'd be really cool if we could expand this into a bigger monitoring and dashboard setup. Just some ideas off the top of my head; these are some places we can start:

1. Wire up Sentry
2. Make/api/health/ actually do health checks rather than be hardcoded
3. Some uptime monitoring 
4. Add support for @vercel/speed-insights.
What do you think @vikram-s-narayan ?

AlvinKuruvilla (approved):

vikram-s-narayan:
> Nice, this looks good to me. It'd be really cool if we could expand this into a bigger monitoring and dashboard setup. Just some ideas off the top of my head; these are some places we can start:
> 
> 1. Wire up Sentry
> 2. Make/api/health/ actually do health checks rather than be hardcoded
> 3. Some uptime monitoring
> 4. Add support for @vercel/speed-insights.
>    What do you think @vikram-s-narayan ?

Sounds awesome! We can start creating issues and implementing :) 

Sidenote: 
I looked up scarf.sh and deferred implementing it because it uses tracking pixels and seems to be making inferences about users. So I was not sure if this would be annoying to some privacy conscious users. Their other technology requires us to put up our docker images in ghcr and will track downloads ... so that may be something worth implementing.

PR #515, opened 2026-09-10
fix(aws-pack): use the official Amazon Bedrock icon

AlvinKuruvilla:
`bedrock.svg` was a hand-drawn approximation; every other icon in the pack is an official AWS Architecture Icon. Replaced with `Arch_Amazon-Bedrock_64`, XML prolog stripped to match the rest of the set — these are injected as HTML rather than parsed as XML.

Needs a pack re-import (`manage.py seed --force`) to reach an existing database; `icon_svg` is populated at import.


vikram-s-narayan (approved):
This is awesome.

PR #517, opened 2026-09-11
ci: mark issues filed outside the issue templates

AlvinKuruvilla:
Tags them with one fixed marker label, `needs-triage`, already created on the repo. Never a category label.

Both issue templates auto-apply a label, so an issue with no labels came in through the blank template or `gh issue create`. 25 of 113 open issues carry none.

`mark` adds `needs-triage` on any issue event when the issue has no labels, and removes it once a real label lands. `sweep` does the same across all open issues on `workflow_dispatch`, for the existing backlog.

This does not guess `bug` vs `enhancement`, it simply adds the `needs-triage` label.

Verified: YAML parses, shellcheck clean, decision logic tested offline against real and synthetic label sets. Not executed — `workflow_dispatch` doesn't register until this is on `main`.

Look at: `mark` and `sweep` are in different concurrency groups, so a sweep can race a concurrent issue event. Adds are idempotent; sweep-add racing mark-remove is reachable.

### DIFF
--- a/.github/workflows/triage-marker.yml
+++ b/.github/workflows/triage-marker.yml
@@ -1,0 +1,106 @@
+name: Issue Triage Marker
+
+# All issue templates, except the blank one, auto-apply a label (bug_report.yml -> `bug`,
+# feature_request.yml -> `enhancement`), so an issue carrying no labels at all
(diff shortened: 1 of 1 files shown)


vikram-s-narayan (approved):
Nice one!

Issue #520, opened 2026-09-12
Canvas table component


AlvinKuruvilla:
A table node on the DFD canvas (`frontend/src/features/dfd-editor/components/nodes/TableNode.tsx`), registered as the `table` node type alongside the sticky note. It carries no DFD semantics: it annotates a diagram and is excluded from threat analysis.

Initial implementation is on `feat/canvas-table-node`.

## Done

- [x] `table` node type, registered in all three node registries
- [x] Text cells, double-click to edit, Tab / Shift-Tab / Enter to move between them
- [x] Column and row resize by dragging dividers, including the outer edges
- [x] Grid picker on the palette entry — hover to size, plus an input mode for sizes the 8×8 grid can't express
- [x] Row and column grips along the top and left edges, click to select
- [x] Right-click any cell or grip: insert row above/below, insert column left/right, delete row, delete column
- [x] Delete key removes the selected row or column
- [x] Corner handles scale the whole table; shift constrains to a uniform factor and scales cell text with it
- [x] Column/row count steppers and a header-row toggle in the edit panel
- [x] Renders in PNG and SVG export, and in the read-only viewer

## Planned

Ordered by evidence rather than by effort. The first two are what it takes to reproduce a table like the ones people build in Miro today.

- [ ] **Text wrapping in cells.** Currently clipped with an ellipsis. Follow Miro: a row's height becomes a *minimum* that content can grow past, so dragging a divider sets a floor rather than a fixed height.
- [ ] **Cell fill colour.** Only the grey header-row flag exists now. Needed for tinting header cells or marking a column.
- [ ] **Paste and copy tabular data** (markdown / CSV / TSV).
- [ ] **Merged cells.** Miro shipped this as a headline feature; jgraph/drawio#1650 has an issue for it.
- [ ] **Drag-to-reorder rows and columns** from the grips, which already exist as selection targets.
- [ ] **Multi-cell selection** and operations across a range.
- [ ] **Per-cell text alignment.**

File individual issues for these as they get picked up, rather than up front — this list is the context, not eight stubs.

## Open design questions

**Merged cells vs. cells holding diagram nodes.** These interact, and the data model already leans one way: a cell is `{ text }` rather than a bare string specifically so it can gain a field pointing at a child node later, giving draw.io-style container tables where a Process sits in a cell and edges connect to it. Merged cells want a span on the same object. Worth deciding together before either is built, since a cell that both spans and contains is a different thing again.

[...]

PR #521, opened 2026-09-12
feat(dfd-editor): add a table node to the canvas

AlvinKuruvilla:
Part of #520.

## Problem

There is no way to lay out tabular content on a DFD canvas — a PMV/context matrix, a STRIDE-per-element grid, a pros/cons comparison next to the diagram it belongs to. Today that means a screenshot, a sticky note, or leaving for another tool.

It is also the single most-requested missing feature in comparable editors: excalidraw#4847 has been open since 2022 with 147 comments, several of them people saying they moved to Lucid, Whimsical or Microsoft Whiteboard over this one gap.

## What this does

Adds a `table` node type. It is an annotation, like the sticky note: no DFD semantics, excluded from threat analysis.

- Text cells that wrap, edited in place, with Tab / Shift-Tab / Enter moving between them and Shift-Enter inserting a line break
- Drag any column or row divider to resize, including the outer edges. A row's stored height is a floor rather than a fixed height, so a row grows when its text needs more — Miro's rule
- Clicking **Table** in the component panel opens a grid picker — hover to size, or switch to an input mode that keeps the grid as a live preview and reaches sizes the 8×8 grid cannot express. Dragging the palette entry still drops a default 3×3
- Grips along the top and left edges select a row or column; right-clicking any cell or grip offers insert above/below/left/right and delete; Delete removes the selected row or column
- Corner handles scale the whole table. Shift constrains to a uniform factor and scales cell text with it, matching Miro
- Column/row steppers and a header-row toggle in the edit panel
- Renders in PNG and SVG export and in the read-only viewer, since `export-diagram-image.ts` runs `html-to-image` over the live DOM

## Where the code is

| file | what |
|---|---|
| `components/nodes/TableNode.tsx` | the node: grid, cells, editing, selection, context menu |
| `hooks/useTableDrag.ts` | both pointer gestures — divider resize and corner scale |
| `components/nodes/table-chrome.tsx` | geometry constants, the row/column grip, the small action buttons |
| `types/diagram.ts` | `TableNodeData` and the pure insert/remove/resize functions |

`TableNode` started at 770 lines against 249 for the next largest file in `nodes/`, and is now 465. Some of the remainder is intrinsic — every other node there renders a shape with a label, this one does grid layout, in-cell editing, keyboard navigation, two pointer gestures, a selection model and a context menu.

The cell and its context menu stay inline deliberately: extracting them costs eight props to move a hundred lines, and that contract reads worse than the version it replaces.

## Alternatives rejected

**`NodeResizer` for the corner handles**, as `StickyNoteNode` uses. It wants to own `style.width`/`height`, and this table's size is derived from its column widths and row heights. Two owners of one dimension start disagreeing, so the corner drag is hand-rolled and writes the widths instead.

**A `ResizeObserver` feeding real row heights back to absolutely positioned overlays.** Once a row can grow past its stored height, any control positioned from a running sum of stored heights drifts off the row it belongs to. Measuring keeps that arithmetic and adds a frame where the overlays are stale. Instead the grips and dividers are grid items in the same tracks as the cells, pushing their visible parts outside with `bottom: 100%` / `right: 100%` — which deletes the arithmetic and makes the misalignment impossible rather than corrected.

**Free-text column/row counts in the edit panel.** Committing per keystroke walks every prefix of what is typed, so 3 → 12 would commit `1` first and truncate nine columns of content. Committing on blur instead means the canvas shows nothing until focus leaves the field. Steppers have neither problem. The grid picker *does* use free text, because nothing exists yet to truncate.

**A cap on table size.** Deliberately unbounded. The cost is real and unpaid for: no virtualization, so a 20×200 table is 4000 DOM nodes the canvas walks on every pan and zoom. If that bites, the fix is virtualizing the cell grid.

## Worth reviewing hardest

- **Zoom correction**, `useTableDrag.ts`. Both gestures divide pointer deltas by `getZoom()`, because sizes are stored in diagram coordinates. Get this wrong and dragging at 50% zoom moves half as far as the cursor — it is what most of the draw.io column-resize bug list is about.
- **The three node registries.** `nodeTypes`, `canvasNodeTypes` and `guestNodeTypes` now assert exhaustiveness with `satisfies Record<DiagramNodeType, unknown>`. Without it React Flow silently substitutes its default node for an unregistered type, which renders as a small box containing the label and reads as a broken component rather than a missing entry. Not hypothetical — it is defect 1 below.
- **Grid placement**, `TableNode.tsx`. Every cell, grip and divider is placed explicitly. A column divider spans `1 / -1` of its column, so between them the overlays claim every defined cell; anything left to auto-placement lands in implicit rows below the table. Defect 2 below.
- **Double-click propagation.** A cell's `onDoubleClick` stops propagation so it never reaches `handleNodeDoubleClick`, which sets `isInlineEditing` on the node. That flag is the fresh-drop hook meaning "caret in the first cell", and without the guard every double-click bounced editing back to cell 1.
- **Two set-during-render blocks** in `TableNode`, seeding the caret and clearing the axis selection. Both derive local state from a changed prop. They were `useEffect`s that had been violating `react-hooks/set-state-in-effect` all along — extracting a second component into the file stopped the React Compiler lint bailing out and it started reporting them.

## Verification

```
npx tsc -b        # clean
npm run build     # clean
npx eslint <every changed file>   # clean
```

Each commit typechecks on its own, checked by building at each SHA. Pre-existing and untouched: one `react-hooks/set-state-in-effect` error at `DFDEditor.tsx:134` and several in the guest-editor tree, all in code this branch does not modify.

This frontend has no test tooling — no vitest, jest or testing-library — so the rest is manual. It has been exercised in the editor, which found and fixed three defects that compiling could not see:

1. The table rendered as a single small box. `guestNodeTypes` had no `table` entry, so React Flow substituted its default node.
2. Rows collapsed to a thin strip under an empty box. The overlays' explicit grid placement claimed every defined cell, pushing the auto-placed cells into implicit rows.
3. Wrapping only took effect on blur. The editing textarea was fixed at one line and scrolled internally, so the row had nothing to grow around until the plain span took over.

Not individually confirmed: divider and corner drags at a non-100% zoom, shift-uniform scaling, the context menu on every entry, and Delete on a selected axis. The refactor landed after that browser pass, so it is compile-verified only.

## Not included

Cell fill colour, markdown/CSV paste and copy, merged cells, drag-to-reorder, multi-cell selection. All tracked with rationale and ordering in #520.

## Compatibility

No breaking changes. `fontSize` on the node data is optional so tables saved before it keep rendering at the previous size, and cells are objects rather than bare strings from the start so that a cell can later hold a diagram node without migrating saved diagrams.

Adds one dependency: `@radix-ui/react-context-menu`, alongside the 15 Radix packages already present.


vikram-s-narayan (approved):

vikram-s-narayan:
@AlvinKuruvilla - this is a superbly executed feature.

PR #551, opened 2026-09-15
feat(compose): rootless Podman overlay for local dev

JJediny:
## Why this matters (plain language)

On locked-down enterprise or federal workstations, the developer often can't get `newuidmap`/`newgidmap` privileges required to run rootful Docker or a rootful Podman machine. That leaves them unable to run the standard `docker-compose.yml` locally — Postgres in particular fails on volume ownership. This PR adds a small, opt-in overlay that maps the Postgres container's UID to the host user so the shared stack Just Works under rootless Podman, without changing anything for the Docker-based flow.

## Value

- Widens who can contribute to Precogly (rootless Podman on managed laptops now works).
- Zero behavior change for the existing Docker workflow — the overlay is only applied when the developer opts in via `-f podman-compose.override.yml`.

## What this changes

- Adds `podman-compose.override.yml` that layers on top of `docker-compose.yml`:
  - `db.userns_mode` maps Postgres UID 70 to the host user so volume chowns succeed under rootless Podman.
  - A fully qualified Postgres image reference (Podman is strict about this).
- All other services and settings come from `docker-compose.yml` unchanged.
- Documents the one-time host configuration and overlay invocation in `docs/getting-started/local-podman.md`, registered in the mkdocs nav.

## Independence

- Local-dev only. No runtime code, no schema, no API surface. Does not depend on and is not depended on by any other open PR.

## Testing

- Verified the overlay boots the stack under rootless Podman on a host without subuid/subgid delegation.
- Docker-based `docker compose up` unchanged (overlay is not loaded by default).


vikram-s-narayan:
@JJediny - Thanks for putting this together! 

Quick question: since local-podman.md already shows the full overlay YAML inline, do we need podman-compose.override.yml checked into the repo? Podman users could create it locally by following the guide. That way we avoid adding a file to the tree that most contributors won't recognize. What do you think?

docs/getting-started/local-podman.md
AlvinKuruvilla, line 1: We should specify the minimum supported Podman version we want users to use.

docs/getting-started/local-podman.md
```diff
@@ -0,0 +1,140 @@
...
+```toml
+[containers]
+default_sysctls = []
+```
+
+Apply the new storage settings (this wipes any existing rootless images and
+containers):
+
+```bash
+podman system reset --force
```
AlvinKuruvilla, line 71: Is there a reason to use --force here? Let's be specific about what the command implies. If we want to offer --force as an option, we can, but we need to give users the full text on its unintended side effects.

podman-compose.override.yml
```diff
@@ -0,0 +1,20 @@
...
+# domain-account uid falls outside /etc/subuid). Layer it on top with:
+#
+#   podman-compose -f docker-compose.yml -f podman-compose.override.yml up -d
+#
+# Docker users do not need this file. See docs/getting-started/local-podman.md
+# for the one-time host configuration this expects.
+services:
+  db:
+    # Fully qualified so rootless Podman does not prompt for a registry.
+    image: docker.io/library/postgres:16-alpine
```
AlvinKuruvilla, line 15: Pinning the image here means the overlay owns the Postgres version for Podman users — an override replaces the scalar; it doesn't annotate it. Bump docker-compose.yml:3 to 17, and Podman users silently stay on 16.

Qualify it in docker-compose.yml instead, and this file is just userns_mode. Docker resolves postgres and docker.io/library/postgres to the same reference, so nothing changes for Docker users.

AlvinKuruvilla (changes requested):
Good work, this seems pretty good, although I'm not very familiar with podman. Just a couple of things I think worth doing

AlvinKuruvilla:
I agree with @vikram-s-narayan, I'm not super familiar with Podman, but if this file is something that will be regenerated, it doesn't seem worth it to have it in the tree

JJediny:
> @JJediny - Thanks for putting this together!
> 
> Quick question: since local-podman.md already shows the full overlay YAML inline, do we need podman-compose.override.yml checked into the repo? Podman users could create it locally by following the guide. That way we avoid adding a file to the tree that most contributors won't recognize. What do you think?

do we need podman-compose.override.yml checked into the repo? Yes - ideally to work out of the box with podman it would make sense to have it present and tracked. Updated now to ONLY include the userspace permissions, everything else is inherited from the docker compose file.

https://podman.io/
https://podman-desktop.io/

AlvinKuruvilla (approved):
Thanks for the changes, LGTM

PR #553, opened 2026-09-17
feat(dfd-editor): select and fill table cells

AlvinKuruvilla:
Canvas tables could only carry text. This adds a cell selection and a colour
fill on top of it, so a row can be marked as the risky one — the first thing
anyone reaches for in a threat model.

Builds on the table node from #, already on `main`.

## What's here

Six commits, each self-contained:

| commit | |
|---|---|
| `chore(ui)` | context menu submenu primitives, matching shadcn's canonical output |
| `fix(dfd-editor)` | a pre-existing focus bug, see below |
| `feat(dfd-editor)` | select cells — click, drag, row/column grips |
| `feat(dfd-editor)` | fill them with a colour |
| `feat(dfd-editor)` | arrow keys move the selection, shift-arrow extends it |
| `feat(dfd-editor)` | the picker shows what the selection already carries |

## Decisions, and what was rejected

**Two-stage drag.** An unselected table drags from anywhere in its body like
every other node; once selected, a press on a cell belongs to the table and the
table moves from its grips. The alternative was a dedicated always-visible drag
border, which costs an affordance on every table and breaks "grab the middle and
move it" outright. This is the Miro rule. It is the change most likely to feel
wrong in the hand, so it is the thing to try first.

`nodrag` on the cells is what enforces it — not the `stopPropagation` in the
hook. React Flow drags from a native listener on the node wrapper, which has
already run by the time a React handler on a descendant fires. The first version
used `stopPropagation` alone and the table moved along with the selection drag.

**Fill is stored per cell**, not per row or column. So a row inserted into a
filled column arrives unfilled, which is visible. The alternative was resolving
cell over column over row on every render and giving "clear this cell" two
meanings.

**The eight colours** are the intersection of Notion's, Google Sheets' and
Airtable's palettes, checked 2026-09-16. Brown is Notion-only; cyan and teal are
absent from Notion; these eight are in all three. Five of the values are already
the sticky note tints, so a yellow cell matches a yellow sticky note beside it —
`TABLE_FILL_COLORS` records that the two sets have to move together, and that
they are frozen Tailwind v3 values while the project runs v4.

**Gray is slate-200, not slate-100.** The header row paints itself slate-100, so
a slate-100 gray would be invisible on row 0 and would read as a second header
row anywhere else.

**No shift-click to extend**, though every spreadsheet has one. Shift is React
Flow's `selectionKeyCode` and a pointerdown under it never reaches the cell;
measured on the guest canvas, Alt and Meta get through and Shift does not.
Keydowns are untouched, which is why shift-*arrow* works. Restoring shift-click
means `selectionKeyCode={null}` on the ReactFlow in `DFDEditor` and
`GuestDFDEditor`, whose only cost is losing selection-box-from-on-top-of-a-node
— measured, not assumed: Shift does nothing else on that canvas. Left out
deliberately; `useTableSelection.ts` records the trade.

## The bug fix

`fix(dfd-editor): keep the caret in a table cell opened after the menu closes`
predates this branch. Using the cell context menu and then double-clicking a
cell to type dropped the caret ~400ms later, because Radix returns focus to the
trigger once the close animation finishes. The insert/delete menu had the same
problem; fill-then-type just makes it easy to hit.

## Verification

- `npx playwright test` — 28 passed, on this branch rebased onto current `main`.
  23 are new, in `frontend/src/features/dfd-editor/__tests__/e2e/table-node.spec.ts`,
  driven through the guest editor because it runs with no backend.
- `npx tsc --noEmit` clean at every commit in the series, not only the tip.
- `npx eslint` clean on every file touched. Pre-existing errors in
  `useDiagramState.ts` and `useParentRelationships.ts` are untouched.
- The gesture behaviour is covered by e2e rather than unit tests because whether
  a press lands on the table or on React Flow's drag depends on real event
  propagation, and nothing short of a browser answers it.

Note for anyone running these locally: Playwright had no browser installed in my
checkout, so `npx playwright install chromium` was needed first.

## Look hardest at

- The two-stage drag, as above.
- `useTableSelection.ts` hit-tests the DOM during a drag rather than measuring
  tracks. A row's rendered height is not knowable from the data — `minmax(h,
  auto)` lets it grow to fit wrapped text — so there is no sum to walk. It uses
  `elementsFromPoint` because the divider drag strips overlay the cells near
  every edge.

## Left out

- Deleting several rows or columns at once. An arrow key clears the grip's
  `axis` claim, so Delete stops being able to remove a row or column once the
  selection has moved or widened.
- Merged cells, per-cell text alignment and borders. The selection is the piece
  they all need and it is here; none of them is started.


AlvinKuruvilla:
#554 (merged cells) branches off this one and is blocked on it landing. It carries the header-spanning that a STRIDE-per-interaction worksheet needs, and it builds directly on the cell selection here — `mergeTableCells` takes a `TableCellRange` and the menu acts on `selectedRange`.

Worth reviewing this one on its own terms first; #554 is a draft until this merges.

vikram-s-narayan (approved):
LGTM

vikram-s-narayan:
@AlvinKuruvilla  - this is such a cool feature! It's a real pleasure to play around with adding different colors to different cells! Also love the smoothness with which we can move from cell to cell with the arrow keys!

PR #554, opened 2026-09-17
feat(dfd-editor): merge table cells and STRIDE worksheet template

AlvinKuruvilla:
Three things that belong together: merging cells, the worksheet that needed
merging, and a context-menu bug the cell selection in #553 shipped with.

Select a range and merge it, so a table can carry a header spanning the columns
under it. That is what a STRIDE-per-interaction worksheet needs (#520): a
component name over its Mitigations/Weaknesses pair, which is the layout Vikram
posted and the reason this exists. That worksheet now ships as a template pack
rather than living only in my checkout.

Screenshot of the rebuilt worksheet is at `frontend/tmp/stride-merged.png` in my
checkout if you want it dropped in here; I can't attach images from the CLI.

## Behaviour, and where it comes from

Checked against the tools on 2026-09-16 rather than guessed.

**Interaction is Miro's** — select cells, *Merge cells* on the context menu, and
they can be unmerged at any time.

**Content is Excel's rule** — the anchor keeps its text, the rest are emptied.
[Microsoft](https://support.microsoft.com/en-us/excel/get-started/merge-and-unmerge-cells-in-excel):
"the contents of only one cell (the upper-left cell…) appear in the merged cell.
The contents of the other cells that you merge are deleted." Sheets is the same.

**No confirmation, unlike both of them.** They warn because in Excel and Sheets
the discarded text is gone for good — Microsoft's doc is explicit that unmerge
does not bring it back. Undo restores it here, and the table already discards
cells without asking when a row or column is removed, with the panel saying so:
*"Removing a column or row discards its cells. Undo restores them."*

**An insert or delete through a merge unmerges it.** Neither spreadsheet does
anything better. Excel expands the selection to the merge's full extent, so
inserting one row inside a ten-row merge inserts
[ten](https://learn.microsoft.com/en-us/answers/questions/5033930/insert-columns-rows-where-there-are-merged-cells);
Sheets often refuses the insert outright. In both, the documented workaround is
unmerge → edit → re-merge, which is what this does for you. Notion does not
support merging at all.

## Data model

Spans live on the anchor as `colSpan`/`rowSpan`, absent meaning one, so a diagram
saved before merging existed is already correct. The cells a merge swallows stay
in the array, keeping the grid rectangular so every index into it still means
what it did. Which cells are covered is **derived** from the anchors, not
flagged — a flag and the span it mirrors are two things that can disagree.

This also answers the open question in #520. A span composes with a later field
pointing at a child node, so a cell that both spans and contains is just an
anchor carrying both. The two features do not need settling together.

## The worksheet

`libraries/packs/worksheets/stride-worksheets`, one template: **STRIDE per
Interaction**. Two elements and the flow between them, a row per STRIDE
category, a Mitigations/Weaknesses pair per element.

**Greyed cells are semantic, not styling** — the categories that do not apply to
that element type, from Vikram's heuristic in #520 (external entity: T, I, D, E;
data store: S, E; data flow: S, R, E; process: none). Grey therefore means "not
applicable" throughout the worksheet, which is why the header rows are yellow: a
grey header would read as a struck-out row.

**`pack_type: template`, not `full`.** It carries no components, threats or
countermeasures. A worksheet is a method rather than a technology domain, so it
has nothing to contribute to the suggestion engine and nothing to join against a
taxonomy — it is complete as it stands, not an unfinished pack.

## Template packs are offered without being connected

`DFDTemplatesLibraryViewSet.get_queryset` gated every template on its source
pack being connected to the threat model, so the `component_ref`s in it resolve;
an AWS template inserted without the AWS pack yields components pointing at
nothing. A template pack ships no components, so a connection has nothing to
supply, and gating it would make a facilitator connect a STRIDE worksheet to
their model as though it described their stack. Those packs are now exempt.

I traced the path from pack YAML to canvas and called it a passthrough before
finding this, which was wrong — the gate sits in a queryset upstream of every
hop I looked at. It only turned up by inserting the template against a running
stack.

## Bugs fixed here

**Merged cells were not clickable through the middle.** Divider drag strips span
the whole table, so the strip for the boundary between two merged rows lay across
the centre of the merged cell — exactly where a pointer lands — and started a row
resize instead. Dividers are now emitted in runs that skip any boundary a merge
crosses. Found by a test timing out on `cursor-row-resize … intercepts pointer
events`; it would have shipped otherwise.

**Arrow keys walked into merged blocks.** Stepping right from a two-wide merge
landed on a covered cell, which renders nothing, so the selection appeared to
vanish. The covered map carries each cell's anchor, and `moveFocus` steps out of
the whole block rather than one cell at a time.

**The context menu would not close over a selected table** — this one shipped in
#553. A selected table swallowed the pointerdown on its cells, so an open menu
stayed open until a click landed outside the table; an unselected one was fine,
since only a selected table takes the pointer at all. React's `stopPropagation`
stops the native event at the root container, and Radix dismisses layers from a
`pointerdown` listener on `document` above it. Nothing needed the stop: `nodrag`
is what keeps a press from dragging the node, and no ancestor of a cell handles
`pointerdown` (@xyflow/react 12.10.0 — its only React-level one is on `Handle`, a
descendant, and the pane's is capture-phase).

## Verification

- `npx playwright test` — 36 passed, including a regression test for the menu
  that fails on the previous commit (`toBeHidden` → received visible).
- `manage.py test apps.diagrams.tests.test_template_pack_visibility` — 2 passed,
  covering both halves: the template pack is offered unconnected, and every other
  pack is still gated.
- `npx tsc --noEmit` and `npx eslint` clean on every file touched.
- Inserted the worksheet from the template browser against a running stack, with
  the pack imported and deliberately not connected to any threat model — the
  state a real user is in.

## Look hardest at

- `expandRangeOverMerges`. Merging a range that cuts an existing merge would
  leave two anchors claiming one cell, which nothing downstream can render. It
  grows the range to swallow whole merges, iterated to a fixed point because
  widening to absorb one can bring the edge against another.
- The queryset relaxation. Widening it further would break technology templates
  quietly, which is why the second test matters more than the first.
- The divider runs, which are the least obvious part of the diff and exist only
  because of the click-stealing bug above.

## Left out

- Merging is not offered across a selection that is already partly merged in a
  way the expansion cannot resolve; it simply widens instead. No case found where
  that is wrong, but it is the area I would fuzz first.
- The guest editor gets no templates, deliberately, so the worksheet is not
  reachable there.
- Reassigning a worksheet column group to a different element type means
  re-greying its column pair by hand. The greys are keyed to the three types the
  template ships with.


vikram-s-narayan:
merged #553

AlvinKuruvilla:
Superseded by #556, which is the same four commits from a branch in this repo rather than my fork. Closing so there is one PR to review.

vikram-s-narayan (approved):
This is super cool!

AlvinKuruvilla:
Thanks, give it a try see if I missed anything or if you have ideas to extend its functionality

